PR #1, opened 2017-08-04
Refactoring and benchmark state using tumbling windows

pnowojski:
@NicoK please take a look

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/1)


src/main/java/org/apache/flink/benchmark/functions/IntegerLongSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void run(SourceContext<Record> ctx) throws Exception {
+        long counter = 0;
+
+        while (running) {
+            synchronized (ctx.getCheckpointLock()) {
+                ctx.collectWithTimestamp(Record.of((int) (counter % numberOfKeys), counter), counter);
+                counter++;
+                if (counter >= numberOfElements) {
```
NicoK, line 50: this doesn't need to be inside the checkpoint lock, does it?

src/main/java/org/apache/flink/benchmark/functions/MultiplyIntLongByTwo.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.flink.benchmark.functions;
+
+import org.apache.flink.api.common.functions.MapFunction;
+
+public class MultiplyIntLongByTwo implements MapFunction<IntegerLongSource.Record, IntegerLongSource.Record> {
+    @Override
+    public IntegerLongSource.Record map(IntegerLongSource.Record value) throws Exception {
```
NicoK: let's call the parameter `record` so that we don't have `value.value`?

src/main/java/org/apache/flink/benchmark/functions/IntegerLongSource.java
```diff
@@ -0,0 +1,61 @@
+package org.apache.flink.benchmark.functions;
+
+import org.apache.flink.streaming.api.functions.source.RichParallelSourceFunction;
+
+public class IntegerLongSource extends RichParallelSourceFunction<IntegerLongSource.Record> {
+    public static class Record {
```
NicoK: should we make this `final`?

src/main/java/org/apache/flink/benchmark/StateBackendBenchmark.java
```diff
@@ -91,7 +93,7 @@ public static void main(String[] args)
 
 	@Benchmark
 	public void stateBackends(StateBackendContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, EventTimeSessionWindows.withGap(Time.seconds(500)));
+		IntLongApplications.reduceWithWindow(context.source, TumblingEventTimeWindows.of(Time.seconds(10_000)));
```
NicoK: These are probably only (mostly?) faster, because tumbling windows don't stress the state back-end that much with their own state (they only ever add values and then aggregate at the end) - session windows may constantly merge sessions and thus state as well. So I'm not sure, that by this change, you are actually getting the opposite, i.e. a higher part of state-unrelated overhead...(at least indirectly)

src/main/java/org/apache/flink/benchmark/RocksStateBackendBenchmark.java
```diff
@@ -0,0 +1,76 @@
...
+
+	@Benchmark
+	public void stateBackends(RocksStateBackendContext context) throws Exception {
+		IntLongApplications.reduceWithWindow(context.source, TumblingEventTimeWindows.of(Time.seconds(10_000)));
+		context.execute();
+	}
+
+	@State(Thread)
+	public static class RocksStateBackendContext extends StateBackendContext {
+		@Param({"ROCKS", "ROCKS_INC"})
```
NicoK, line 63: should we also have `MEMORY` here so that we can compare?

NicoK (changes requested):
Looks good in general with some questions inline.
I did not verify the changes functionally but I suppose, you did that :)

src/main/java/org/apache/flink/benchmark/functions/IntegerLongSource.java
pnowojski: I guess not, but does it matter? It makes code a little bit cleaner and this if condition is extremely in expensive.

src/main/java/org/apache/flink/benchmark/StateBackendBenchmark.java
pnowojski: The idea is that session windows are already so slow, that they can hide most of the cost of using rocks db. That's why @StephanEwen proposed to change it to "cheapest" windows.

src/main/java/org/apache/flink/benchmark/RocksStateBackendBenchmark.java
pnowojski: We don't need to, because `MEMORY` is already being tested by separate benchmark. Adding it here would just be redundant. 

Regardless of that, comparison between `ROCKS` and `MEMORY` is actually not that important, besides some general idea "how much slower it is" and some pure curiosity. For actual daily usage and productional value, you care only whether your recent changes slowed down or improved either `ROCKS` or `MEMORY`.

pnowojski: Ops, and there are benchmarks for `MEMORY` `FS` etc, but there are in different class `MemoryStateBackendBenchmark`. I have split them, because they had to run on different number of records per invocation :( `ROCKS` was way too slow for `7 000 000` records and `MEMORY` is way too fast for `2 000 000` records :(

src/main/java/org/apache/flink/benchmark/functions/IntegerLongSource.java
NicoK: you're running micro benchmarks, so it probably does matter :P - you could go with the following (which I'd consider even cleaner)
```
while (running && counter >= numberOfElements) { ... }
running = false;
```

src/main/java/org/apache/flink/benchmark/RocksStateBackendBenchmark.java
NicoK: ok, if we do not need the comparison, than this is ok

pnowojski:
Merged

PR #2, opened 2017-08-31
Benchmarks for PreAggregationOperator

pnowojski:
Most of the commits are from #1 

Added PreAggregationOperator benchmarks and trimmed window sizes in other benchmarks to more reasonable values. Trimming will probably brake benchmark results in codespeed.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/WindowBenchmarks.java
+++ b/src/main/java/org/apache/flink/benchmark/WindowBenchmarks.java
@@ -42,7 +42,11 @@ import java.io.IOException;
 @OperationsPerInvocation(value = WindowBenchmarks.RECORDS_PER_INVOCATION)
 public class WindowBenchmarks extends BenchmarkBase {
 
+	// IntegerLongSource produces records (logically) at rate 1 record per 1 ms, so 1_000_000 records will span over
+	// 1_000_000 milliseconds.
 	public static final int RECORDS_PER_INVOCATION = 7_000_000;
+	public static final Time WINDOW_LENGTH = Time.milliseconds(100_000);
+	public static final Time SLIDING_WINDOW_SLIDE = Time.milliseconds(10_000);
 
 	public static void main(String[] args)
 			throws RunnerException {
@@ -62,19 +66,19 @@ public class WindowBenchmarks extends BenchmarkBase {
 
 	@Benchmark
 	public void tumblingWindow(TimeWindowContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, TumblingEventTimeWindows.of(Time.seconds(10_000)));
+		IntLongApplications.reduceWithWindow(context.source, TumblingEventTimeWindows.of(WINDOW_LENGTH));
 		context.execute();
 	}
 
 	@Benchmark
 	public void slidingWindow(TimeWindowContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, SlidingEventTimeWindows.of(Time.seconds(10_000), Time.seconds(1000)));
+		IntLongApplications.reduceWithWindow(context.source, SlidingEventTimeWindows.of(WINDOW_LENGTH, SLIDING_WINDOW_SLIDE));
 		context.execute();
 	}
 
 	@Benchmark
 	public void sessionWindow(TimeWindowContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, EventTimeSessionWindows.withGap(Time.seconds(500)));
+		IntLongApplications.reduceWithWindow(context.source, EventTimeSessionWindows.withGap(WINDOW_LENGTH));
 		context.execute();
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/flink/benchmark/WindowBenchmarks.java
```diff
@@ -42,7 +42,11 @@
 @OperationsPerInvocation(value = WindowBenchmarks.RECORDS_PER_INVOCATION)
 public class WindowBenchmarks extends BenchmarkBase {
 
+	// IntegerLongSource produces records (logically) at rate 1 record per 1 ms, so 1_000_000 records will span over
+	// 1_000_000 milliseconds.
 	public static final int RECORDS_PER_INVOCATION = 7_000_000;
+	public static final Time WINDOW_LENGTH = Time.milliseconds(100_000);
+	public static final Time SLIDING_WINDOW_SLIDE = Time.milliseconds(10_000);
```
NicoK, line 49: we should probably add some comments on why we chose these values to know in the future

src/main/java/org/apache/flink/benchmark/WindowBenchmarks.java
```diff
@@ -62,19 +66,25 @@ public void globalWindow(TimeWindowContext context) throws Exception {
...
 	public void slidingWindow(TimeWindowContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, SlidingEventTimeWindows.of(Time.seconds(10_000), Time.seconds(1000)));
+		IntLongApplications.reduceWithWindow(context.source, SlidingEventTimeWindows.of(WINDOW_LENGTH, SLIDING_WINDOW_SLIDE));
 		context.execute();
 	}
 
 	@Benchmark
 	public void sessionWindow(TimeWindowContext context) throws Exception {
-		IntLongApplications.reduceWithWindow(context.source, EventTimeSessionWindows.withGap(Time.seconds(500)));
+		IntLongApplications.reduceWithWindow(context.source, EventTimeSessionWindows.withGap(WINDOW_LENGTH));
```
NicoK, line 81: Strictly speaking, here, we are talking about the gap between sessions, not the `WINDOW_LENGTH`. Would it rather make sense to separate this?

src/main/java/org/apache/flink/benchmark/functions/IntLongApplications.java
```diff
@@ -16,4 +26,68 @@
...
+                                windowAssigner,
+                                false))
+                .map(new MapFunction<Tuple3<Integer, TimeWindow, LongAccumulator>, IntegerLongSource.Record>() {
+                    @Override
+                    public IntegerLongSource.Record map(Tuple3<Integer, TimeWindow, LongAccumulator> value) throws Exception {
+                        return new IntegerLongSource.Record(value.f0, value.f2.value);
+                    }
+                })
+                .keyBy(record -> record.key)
+                .window(windowAssigner)
```
NicoK, line 54: I just checked with Stefan that although this `windowAssigner` instance is shared by different operators in the job definition here, they will be different in the actual execution (which is desired). Maybe we should reflect this here as well, though, so have two window assigner instances: one for each use case.

On the other hand... you want the two windows to reflect the same semantics, so it's probably fine.

src/main/java/org/apache/flink/benchmark/functions/IntLongApplications.java
```diff
@@ -16,4 +26,68 @@
...
+        public LongAccumulator() {
+            this(0L);
+        }
+
+        public LongAccumulator(long value) {
+            this.value = value;
+        }
+    }
+
+    private static class RecordRecordVoidAggregateFunction
```
NicoK, line 71: the naming is somewhat strange...where does the second `Record` come from?

src/main/java/org/apache/flink/benchmark/functions/IntegerLongSource.java
```diff
@@ -41,12 +42,16 @@ public IntegerLongSource(int numberOfKeys, long numberOfElements) {
...
-        long counter = 0;
+        final int parallelism = getRuntimeContext().getNumberOfParallelSubtasks();
+        long counter = getRuntimeContext().getIndexOfThisSubtask();
 
         while (running && counter < numberOfElements) {
             synchronized (ctx.getCheckpointLock()) {
                 ctx.collectWithTimestamp(Record.of((int) (counter % numberOfKeys), counter), counter);
-                counter++;
+                counter += parallelism;
+                if ((counter / parallelism) % (1024 * 16) == 0) {
```
NicoK, line 52: Please add a comment which at least explains the reasoning behind these values.

Also, you could simplify the calculation to reduce the impact on the source's performance. I guess the least impact is by using a simple `int emitWatermarkCounter` that starts with the value you'd like for the repetition, decreases every round, and is reset once 0 is reached.

NicoK (changes requested):
Didn't look at the `PreAggregationOperator` use but I guess you know what you're doing there ;) - overall, the rest looks fine, but I'd like a few changes here and there

PR #3, opened 2017-12-07
Execute NetworkBenchmarks from Flink

pnowojski:


src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -62,20 +63,37 @@ public void networkThroughput(MultiEnvironment context) throws Exception {
...
 		@Setup
 		public void setUp() throws Exception {
+			int channels = parseChannels(channelsFlushTimeout);
+			int flushTimeout = parseFlushTimeout(channelsFlushTimeout);
 			super.setUp(writers, channels, flushTimeout);
 		}
 
+		private static int parseFlushTimeout(String channelsFlushTimeout) {
+			String[] parameters = channelsFlushTimeout.split(",");
+			checkState(parameters.length >= 2);
```
NicoK: shouldn't it rather be `checkArgument`?

src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -62,20 +63,37 @@ public void networkThroughput(MultiEnvironment context) throws Exception {
...
-		@Param({"1", "100"})
-		public int flushTimeout = 1;
+		//Do not spam continuous benchmarking with number of writers parameter.
+		//@Param({"1", "4"})
+		public int writers = 4;
 
 		@Setup
 		public void setUp() throws Exception {
+			int channels = parseChannels(channelsFlushTimeout);
+			int flushTimeout = parseFlushTimeout(channelsFlushTimeout);
```
NicoK: At some point, we should do the parsing only once, but imho, this is ok for now.

src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -62,20 +63,37 @@ public void networkThroughput(MultiEnvironment context) throws Exception {
...
 			super.setUp(writers, channels, flushTimeout);
 		}
 
+		private static int parseFlushTimeout(String channelsFlushTimeout) {
+			String[] parameters = channelsFlushTimeout.split(",");
+			checkState(parameters.length >= 2);
+			String flushTimeout = parameters[1];
+
+			checkState(flushTimeout.endsWith("ms"));
```
NicoK, line 87: `checkArgument`?

src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -62,20 +63,37 @@ public void networkThroughput(MultiEnvironment context) throws Exception {
...
+			return Integer.parseInt(flushTimeout.substring(0, flushTimeout.length() - 2));
+		}
+
+		private static int parseChannels(String channelsFlushTimeout) {
+			String[] parameters = channelsFlushTimeout.split(",");
+			checkState(parameters.length >= 1);
```
NicoK, line 93: `checkArgument`?

NicoK (changes requested):
ok, I would just change from `checkState` to `checkArgument` - rest looks ok, basically

src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputBenchmarkExecutor.java
pnowojski: Yes, agree, I'm not proud of it, but I didn't want to over-engineer it at this moment by introducing additional pojo to return both values at once :(

pnowojski:
Thanks for the review! merged

PR #5, opened 2018-06-04
add benchmarks for the keyBy() operation on tuples and arrays

NicoK:


src/main/java/org/apache/flink/benchmark/KeyByBenchmarks.java
```diff
@@ -0,0 +1,137 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+/**
+ * Benchmark for keyBy() on tuples and arrays.
+ */
+public class KeyByBenchmarks extends BenchmarkBase {
+
+	private static final int TUPLE_RECORDS_PER_INVOCATION = 25_000_000;
```
pnowojski: Does those numbers of records must be that high? Other benchmarks are using `7_000_000`.

The lower the number of records, the faster or the more accurate the benchmarks will be. JMH executes as many invocations in a single run as possible as long as in total they do not exceed 1 second. Thus:
- if one invocation is already > 1 sec, smaller number of records => faster benchmarks
- if one invocation is already < 1 sec, smaller number of records => more invocations => more accurate results.

The only reason for larger number of records is if profiling shows that initialising/rump up is taking too much time.

src/main/java/org/apache/flink/benchmark/KeyByBenchmarks.java
```diff
@@ -0,0 +1,137 @@
...
+
+		env.execute();
+	}
+
+	private abstract static class IncreasingBaseSource<T> implements ParallelSourceFunction<T> {
+		private static final long serialVersionUID = 8318018060123048234L;
+		final int numKeys;
+		int remainingEvents;
+
+		IncreasingBaseSource(int numKeys, int numEvents) {
```
pnowojski: order of `int numEvents, int numKeys` is inconsistent

pnowojski (approved):
LGTM % number of records

src/main/java/org/apache/flink/benchmark/KeyByBenchmarks.java
NicoK: Indeed, fewer records also resulted in the same values. For this simple job, however, the ramp-up is a significant portion of the benchmark if the number of records it too small.

PR #7, opened 2018-08-29
add benchmarks for different serialization frameworks

NicoK:
FYI: These are the results with Flink 1.6.0 on my machine:
```
Benchmark                                                       Mode  Cnt    Score    Error   Units
SerializationFrameworkBenchmarks.testAvroForcedSerializer      thrpt   30  115.814 ±  2.714  ops/ms
SerializationFrameworkBenchmarks.testAvroSchemaSerializer      thrpt   30  160.859 ±  4.622  ops/ms
SerializationFrameworkBenchmarks.testKryoSerializer            thrpt   30   94.518 ±  1.939  ops/ms
SerializationFrameworkBenchmarks.testKryoSerializerRegistered  thrpt   30  120.472 ± 12.675  ops/ms
SerializationFrameworkBenchmarks.testPojoSerializer            thrpt   30  234.788 ±  5.263  ops/ms
SerializationFrameworkBenchmarks.testPojoSerializerRegistered  thrpt   30  256.074 ± 16.486  ops/ms
SerializationFrameworkBenchmarks.testRowSerializer             thrpt   30  297.823 ± 20.833  ops/ms
SerializationFrameworkBenchmarks.testTupleSerializer           thrpt   30  313.929 ±  9.551  ops/ms
```

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
+++ b/src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark.functions;
+
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.streaming.api.functions.source.ParallelSourceFunction;
+
+/**
+ * Abstract base class for sources with a defined number of events and a fixed key range.
+ */
+abstract public class BaseSourceWithKeyRange<T> implements ParallelSourceFunction<T> {
+	private static final long serialVersionUID = 8318018060123048234L;
+
+	protected final int numKeys;
+	protected int remainingEvents;
+
+	public BaseSourceWithKeyRange(int numEvents, int numKeys) {
+		this.remainingEvents = numEvents;
+		this.numKeys = numKeys;
+	}
+
+	protected abstract T getElement(int keyId);
+
+	@Override
+	public void run(SourceContext<T> out) {
+		int keyId = 0;
+		while (--remainingEvents >= 0) {
+			T element = getElement(keyId);
+			synchronized (out.getCheckpointLock()) {
+				out.collect(element);
+			}
+			++keyId;
+			if (keyId >= numKeys) {
+				keyId = 0;
+			}
+		}
+	}
+
+	@Override
+	public void cancel() {
+		this.remainingEvents = 0;
+	}
+}
--- a/src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
+++ b/src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
@@ -35,10 +35,15 @@ abstract public class BaseSourceWithKeyRange<T> implements ParallelSourceFunctio
 		this.numKeys = numKeys;
 	}
 
+	protected void init() {
+	}
+
 	protected abstract T getElement(int keyId);
 
 	@Override
 	public void run(SourceContext<T> out) {
+		init();
+
 		int keyId = 0;
 		while (--remainingEvents >= 0) {
 			T element = getElement(keyId);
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
```diff
@@ -0,0 +1,59 @@
...
+
+	@Override
+	public void run(SourceContext<T> out) {
+		int keyId = 0;
+		while (--remainingEvents >= 0) {
+			T element = getElement(keyId);
+			synchronized (out.getCheckpointLock()) {
+				out.collect(element);
+			}
+			++keyId;
```
pnowojski, line 53: I like the spirit of extracting every single instruction out of the lock, but maybe this `getElement ` and `++key` are a bit too much ;)

src/main/java/org/apache/flink/benchmark/SerializationFrameworkBenchmarks.java
```diff
@@ -0,0 +1,451 @@
...
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Arrays;
+
+/**
+ * Benchmark for serializing POJOs and Tuples with different serialization frameworks.
+ */
+public class SerializationFrameworkBenchmarks extends BenchmarkBase {
+
+	private static final int RECORDS_PER_INVOCATION = 100_000;
```
pnowojski: `100_000` records per invocation and judging from the results that you posted one invocation takes ~1 second, right? If you double the value, the results stay the same?

src/main/java/org/apache/flink/benchmark/SerializationFrameworkBenchmarks.java
```diff
@@ -0,0 +1,451 @@
...
+		env.addSource(new PojoSource(RECORDS_PER_INVOCATION, 10))
+				.rebalance()
+				.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkBenchmarks.RECORDS_PER_INVOCATION)
+	public void testPojoSerializerRegistered() throws Exception {
```
pnowojski: why registering is faster?

src/main/java/org/apache/flink/benchmark/SerializationFrameworkBenchmarks.java
```diff
@@ -0,0 +1,451 @@
...
+		env.addSource(new PojoSource(RECORDS_PER_INVOCATION, 10))
+				.rebalance()
+				.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkBenchmarks.RECORDS_PER_INVOCATION)
+	public void testAvroSchemaSerializer() throws Exception {
```
pnowojski: please drop the `test` prefix from the methods. It's redundant with `@Benchmark` annotation and longer names are more problematic for `codespeed` ui.

pnowojski (commented):
More or less the change LGTM.

I have two concerns:
- records that you are serialising are array "heavy", thus arrays overheads could be dominant factor when comparing avro, pojo, row, tuple and kryo
- do we need `testAvroForcedSerializer`, `testKryoSerializer` and `testPojoSerializer`? How likely is that we will brake something for them, while not braking for registered kryo/pojo? (Asking because I would like to limit number of benchmarks)

src/main/java/org/apache/flink/benchmark/functions/BaseSourceWithKeyRange.java
NicoK: I don't see a problem here...both accesses are only from a single thread anyway, so no locking required

src/main/java/org/apache/flink/benchmark/SerializationFrameworkBenchmarks.java
NicoK: I'm assuming, then we also do not use full class names in the serialization format but rather registered tags

src/main/java/org/apache/flink/benchmark/SerializationFrameworkBenchmarks.java
NicoK: true - they don't :(
with `300_000` I get the following results:
```
SerializationFrameworkBenchmarks.testAvroForcedSerializer      thrpt   30  129.923 ±  3.274  ops/ms
SerializationFrameworkBenchmarks.testAvroSchemaSerializer      thrpt   30  216.473 ± 34.747  ops/ms
SerializationFrameworkBenchmarks.testKryoSerializer            thrpt   30   70.413 ±  1.143  ops/ms
SerializationFrameworkBenchmarks.testKryoSerializerRegistered  thrpt   30  104.970 ±  1.541  ops/ms
SerializationFrameworkBenchmarks.testPojoSerializer            thrpt   30  344.357 ± 25.534  ops/ms
SerializationFrameworkBenchmarks.testPojoSerializerRegistered  thrpt   30  344.299 ± 22.487  ops/ms
SerializationFrameworkBenchmarks.testRowSerializer             thrpt   30  416.346 ± 29.662  ops/ms
SerializationFrameworkBenchmarks.testTupleSerializer           thrpt   30  440.312 ±  5.931  ops/ms
```
it's a strange thing that Kryo actually got slower (as the only benchmark!)

NicoK: looks like 300k is a good enough number: 100k and 200k have significantly lower numbers for Pojo, Row, and Tuple while the gain by 400k and 500k is not significant any more (at least on my PC)

NicoK:
done - pushed the changes as agreed

pnowojski (approved):
LGTM

PR #8, opened 2018-08-29
update Flink to 1.7-SNAPSHOT

NicoK:
- also clean up the `pom.xml`


tisonkun:
I've checked out your branch and run it on my local env(macOS 10.13.6) with this command `mvn clean install exec:exec`

With the output as below I think this pull request makes sense.


[benchmark.txt](https://github.com/dataArtisans/flink-benchmarks/files/2343989/benchmark.txt)

pnowojski (approved):

PR #9, opened 2018-09-14
Update license to make it compatible with jmh

pnowojski:
We are using jmh not only in tests, but also in main, thus we probably should change the license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,201 +1,674 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                    GNU GENERAL PUBLIC LICENSE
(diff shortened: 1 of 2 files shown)


NicoK:
I'm not sure this step is actually needed. For sure, if anyone is distributing our (compiled) code along with the JMH jars, the result will be GPL but our code does not have to be, especially since we don't want to provide any binary distribution.

To get misunderstandings out of the way, we may follow this path:
- contribute the essence of the benchmarks to [apache/flink](https://github.com/apache/flink)
- only put the JMH benchmark executors into this project
- make this project GPL

pnowojski:
Updated README.md to reflect/explain above assumption/code structure

PR #12, opened 2019-01-30
add SSL network benchmarks and change overall benchmark set

NicoK:
- remove `1,100ms` and `100,1ms`
- add `100,100ms`
- add the following benchmarks with SSL enabled (both also exist without SSL): `100,100ms,SSL` and `1000,100ms,SSL`


pnowojski (approved):

pnowojski:
LGTM

I agree `1,100ms` and `100,1ms` were not very valuable any more. `100,100ms` is a better benchmark to test for the most common setup and +1 for running the SSL only with the most common setup and with high number of nodes.

PR #13, opened 2019-03-20
[FLINK-11986] [state backend, tests] Add micro benchmark for state operations

carp84:
Currently we already have benchmarks for the whole backend, but none for finer grained state operations, and here we propose to add more benchmarks, including (but not limited to):

* ValueState
  * testPut
  * testGet
* ListState
  * testUpdate
  * testGet
  * testAddAll
* MapState
  * testPut
  * testGet
  * testContains
  * testKeys
  * testValues
  * testEntries
  * testIterator
  * testRemove
  * testPutAll

And we will create benchmark for `HeapKeyedStateBackend` and `RocksDBKeyedStateBackend` separately.

### DIFF
--- a/src/main/java/org/apache/flink/state/benchmark/StateBenchmarkConstants.java
+++ b/src/main/java/org/apache/flink/state/benchmark/StateBenchmarkConstants.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.state.benchmark;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Random;
+
+/**
+ * Constants for state benchmark tests. Also generates random keys/values in advance to avoid
+ * possible affect of using {@link Random#nextLong()}
+ */
+class StateBenchmarkConstants {
+    static final int mapKeyCount = 10;
+    static final int listValueCount = 100;
+    static final int setupKeyCount = 500_000;
+    static final String rootDirName = "benchmark";
+    static final String recoveryDirName = "localRecovery";
+    static final String dbDirName = "dbPath";
+
+    static final ArrayList<Long> mapKeys = new ArrayList<>(mapKeyCount);
+
+    static {
+        for (int i = 0; i < mapKeyCount; i++) {
+            mapKeys.add((long) i);
+        }
+        Collections.shuffle(mapKeys);
+    }
+
+    static final ArrayList<Double> mapValues = new ArrayList<>(mapKeyCount);
+
+    static {
+        Random random = new Random();
+        for (int i = 0; i < mapKeyCount; i++) {
+            mapValues.add(random.nextDouble());
+        }
+        Collections.shuffle(mapValues);
+    }
+
+    static final ArrayList<Long> setupKeys = new ArrayList<>(setupKeyCount);
+
+    static {
+        for (long i = 0; i < setupKeyCount; i++) {
+            setupKeys.add(i);
+        }
+        Collections.shuffle(setupKeys);
+    }
+
+    static final int newKeyCount = 500_000;
+    static final ArrayList<Long> newKeys = new ArrayList<>(newKeyCount);
+
+    static {
+        for (long i = 0; i < newKeyCount; i++) {
+            newKeys.add(i + setupKeyCount);
+        }
+        Collections.shuffle(newKeys);
+    }
+
+    static final int randomValueCount = 1_000_000;
+    static final ArrayList<Long> randomValues = new ArrayList<>(randomValueCount);
+
+    static {
+        for (long i = 0; i < randomValueCount; i++) {
+            randomValues.add(i);
+        }
+        Collections.shuffle(randomValues);
+    }
+}
--- a/README.md
+++ b/README.md
@@ -5,16 +5,32 @@ This repository contains sets of micro benchmarks designed to run on single mach
 their changes. 
 
 The main methods defined in the various classes (test cases) are using [jmh](http://openjdk.java.net/projects/code-tools/jmh/)  micro
-benchmark suite to define runners to execute those test cases. You can either execute 
-whole benchmark suite (which takes ~1hour) at once:
+benchmark suite to define runners to execute those test cases. You can execute the
+default benchmark suite (which takes ~1hour) at once:
 
 ```
 mvn -Dflink.version=1.5.0 clean install exec:exec
 ```
 
-or if you want to execute just one benchmark, the best approach is to execute selected main function manually. 
-For example from your IDE (hint there is a plugin for Intellij IDEA). In that case don't forget about selecting 
-`flink.version`, default value for the property is defined in pom.xml.
+There is also a separate benchmark suit for state backend, and you can execute this suit (which takes ~1hour) using
+below command:
+
+```
+mvn -Dflink.version=1.5.0 clean package exec:exec \
+ -Dexec.executable=java -Dexec.args=-jar target/benchmarks.jar -rf csv org.apache.flink.state.benchmark.*
+```
+
+If you want to execute just one benchmark, the best approach is to execute selected main function manually.
+There're mainly two ways:
+
+1. From your IDE (hint there is a plugin for Intellij IDEA).
+   * In this case don't forget about selecting `flink.version`, default value for the property is defined in pom.xml.
+
+2. From command line, using command like:
+   ```
+   mvn -Dflink.version=1.5.0 clean package exec:exec \
+    -Dexec.executable=java -Dexec.args=-jar target/benchmarks.jar <benchmark_class>
+   ```
 
 ## Code structure
 
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@ under the License.
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<flink.version>1.7-SNAPSHOT</flink.version>
+		<flink.version>1.8-SNAPSHOT</flink.version>
 		<java.version>1.8</java.version>
 		<scala.binary.version>2.11</scala.binary.version>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,36 @@ under the License.
 
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.2.0</version>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+						<configuration>
+							<finalName>benchmarks</finalName>
+							<transformers>
+								<transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+									<mainClass>org.openjdk.jmh.Main</mainClass>
+								</transformer>
+                                                                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                                                                        <resource>reference.conf</resource>
+                                                                </transformer>
+                                                                <!-- The service transformer is needed to merge META-INF/services files -->
+                                                                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                                                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
+                                                                        <projectName>Apache Flink</projectName>
+                                                                </transformer>
+							</transformers>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
--- a/save_jmh_result.py
+++ b/save_jmh_result.py
@@ -28,6 +28,8 @@ parser.add_argument('--environment', dest='environment', required=True)
 parser.add_argument('--dry', dest='dry', action='store_true')
 parser.add_argument('--codespeed', dest='codespeed', default=DEFAULT_CODESPEED_URL,
                     help='codespeed url, default: %s' % DEFAULT_CODESPEED_URL)
+parser.add_argument('--project', dest='project', default="Flink")
+parser.add_argument('--exec', dest='executable', default="Flink")
 
 def readData(args):
     results = []
@@ -58,8 +60,8 @@ def readData(args):
             results.append({
                 'commitid': args.commit,
                 'branch': args.branch,
-                'project': 'Flink',
-                'executable': 'Flink',
+                'project': args.project,
+                'executable': args.executable,
                 'benchmark': name,
                 'environment': args.environment,
                 'lessisbetter': False,
--- a/src/main/java/org/apache/flink/state/benchmark/BackendType.java
+++ b/src/main/java/org/apache/flink/state/benchmark/BackendType.java
@@ -1,0 +1,25 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.state.benchmark;
+
+/**
+ * Enum of backend type.
+ */
+public enum BackendType {
+    HEAP, ROCKSDB
+}
--- a/src/main/java/org/apache/flink/state/benchmark/BackendUtils.java
+++ b/src/main/java/org/apache/flink/state/benchmark/BackendUtils.java
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.state.benchmark;
+
+import org.apache.flink.api.common.ExecutionConfig;
+import org.apache.flink.api.common.JobID;
+import org.apache.flink.api.common.typeutils.base.LongSerializer;
+import org.apache.flink.contrib.streaming.state.PredefinedOptions;
+import org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend;
+import org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackendBuilder;
+import org.apache.flink.contrib.streaming.state.RocksDBStateBackend;
+import org.apache.flink.core.fs.CloseableRegistry;
+import org.apache.flink.metrics.groups.UnregisteredMetricsGroup;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
+import org.apache.flink.runtime.state.AbstractStateBackend;
+import org.apache.flink.runtime.state.KeyGroupRange;
+import org.apache.flink.runtime.state.LocalRecoveryConfig;
+import org.apache.flink.runtime.state.LocalRecoveryDirectoryProviderImpl;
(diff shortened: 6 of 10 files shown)


pnowojski:
Could you paste/link the example full output of the benchmark run? Benchmark scores, including the output of all of the measurement/warm up iterations etc? The things that I would like to know are:
1. total time required to execute those new benchmarks.
2. stability of the results

src/main/java/org/apache/flink/state/benchmark/HeapListStateBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.io.IOException;
+
+/**
+ * Benchmark of list state for heap keyed state backend.
+ */
+public class HeapListStateBenchmark extends ListStateBenchmarkBase{
```
StefanRRichter: Missing space before `{`

src/main/java/org/apache/flink/state/benchmark/ValueStateBenchmarkBase.java
```diff
@@ -0,0 +1,83 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    protected abstract KeyedStateBackend<Long> createKeyedStateBackend() throws Exception;
+
+    @Benchmark
+    public void test1Put() throws IOException {
+        backend.setCurrentKey(random.nextLong(setupKeyCount));
```
StefanRRichter: This seems like we are always moving inside the initial keyspace from the setup, so this mens we never benchmark inserting new keys. Could make sense to also test the cost of adding new keys.

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmarkBase.java
```diff
@@ -0,0 +1,97 @@
...
+    @Benchmark
+    public void test1Update() throws Exception {
+        keyedStateBackend.setCurrentKey(random.nextLong(setupKeyCount));
+        listState.update(Collections.singletonList(random.nextLong()));
+    }
+
+    @Benchmark
+    public void test2Get() throws Exception {
+        keyedStateBackend.setCurrentKey(random.nextLong(setupKeyCount));
+        listState.get();
```
StefanRRichter: I think get can have a lazy implementation in the iterator, so i would suggest to also include doing the full iteration over list values.

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmarkBase.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Benchmark
+    public void test3AddAll() throws Exception {
+        keyedStateBackend.setCurrentKey(random.nextLong(setupKeyCount));
+        listState.addAll(dummyLists);
+    }
```
StefanRRichter: I would suggest to think about adding a read/modify/write test here as well - isolating read and write does not give all insights, e.g. in RocksDB write is a simple append but following reads will get slower as long as they are not merged because they need to do a reconstruction of the list from fragments.

src/main/java/org/apache/flink/state/benchmark/ValueStateBenchmarkBase.java
```diff
@@ -0,0 +1,83 @@
...
+        valueState.update(random.nextLong());
+    }
+
+    @Benchmark
+    public void test2Get() throws IOException {
+        backend.setCurrentKey(random.nextLong(setupKeyCount));
+        valueState.value();
+    }
+}
```
StefanRRichter: In general, I wonder if adding read-modify-write tests would not be valueable for additional insight.

src/main/java/org/apache/flink/state/benchmark/ValueStateBenchmarkBase.java
StefanRRichter: For very fast methods the overhead of creating random values inside the benchmarked code could potentially influence the result. I am not sure if "very fast" applies to heap backend in comparison to `random.nextLong` but we should make sure this does not slow down the benchmark. The alternative with minimized overhead would be an array of precomputed random values read in linear iteration.

src/main/java/org/apache/flink/state/benchmark/MapStateBenchmarkBase.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+import static org.apache.flink.state.benchmark.StateBenchmarkConstants.mapKeyCount;
+import static org.apache.flink.state.benchmark.StateBenchmarkConstants.setupKeyCount;
+
+/**
+ * Base implementation for map state benchmark testing.
+ */
+public abstract class MapStateBenchmarkBase extends BenchmarkBase {
+    private MapState<Long, Double> mapState;
```
StefanRRichter: I think those fields can be `final`

src/main/java/org/apache/flink/state/benchmark/RocksDBValueStateBenchmark.java
```diff
@@ -0,0 +1,60 @@
...
+    public static void main(String[] args) throws RunnerException {
+        Options opt = new OptionsBuilder()
+                .verbosity(VerboseMode.NORMAL)
+                .include(".*" + RocksDBValueStateBenchmark.class.getSimpleName() + ".*")
+                .build();
+
+        new Runner(opt).run();
+    }
+
+
```
StefanRRichter: Two newline

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkConstants.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.flink.state.benchmark;
+
+/**
+ * Constants for state benchmark tests.
+ */
+class StateBenchmarkConstants {
+    static final int mapKeyCount = 10;
+    static final int listValueCount = 100;
+    static final int setupKeyCount = 500_000;
```
StefanRRichter, line 32: I have a question about the choice of values: what are we targeting in the benchmark for heap and rock? For example, for heap are would we expect that random op will be answered from memory or from L2/L3 because the whole dataset can still fit there? Similar question for Rocks, do we measure performance for cached blocks or when hitting disk? Should we target the different alternatives, because the message can be very different - for example what if or Rocks benchmarks look all good but never hit the disk and then there are seeks involved to become terrible for users once they hit the disk?

StefanRRichter (changes requested):
Thanks for the work @carp84, I think this is a very good addition to the performance tests. I had a couple of comments, mostly smaller. I think the most imporant ones are about iterating the list state in get and thinking about adding read-modify-write cycle benchmarks. I also had a question about the targeted scenarios (cache/mem/disk). Furthermore, I wonder if it would make sense to extend the tests in the future to include: timer service performances, check/savepointpoint performances, operational performances with concurrently running checkpoints.

carp84:
> Could you paste/link the example full output of the benchmark run? Benchmark scores, including the output of all of the measurement/warm up iterations etc? The things that I would like to know are:
> 
> 1. total time required to execute those new benchmarks.
> 2. stability of the results

Thanks for review @pnowojski. Let me get the data after resolving Stefan's review comments since possibly the change will affect the result.

src/main/java/org/apache/flink/state/benchmark/ValueStateBenchmarkBase.java
carp84: Agreed. Will change this for all benchmarks.

carp84:
> Furthermore, I wonder if it would make sense to extend the tests in the future to include: timer service performances, check/savepointpoint performances, operational performances with concurrently running checkpoints.

We also have internal benchmarks for timer service and checkpoint performance, and let me do upstreaming one by one (smile). For performance with concurrently running checkpoints I agree we need to add one.

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkConstants.java
carp84: We set the values large enough to trigger disk operations for rocksdb, and confirm this by checking rocksdb log to see whether there's flush/compaction happened. However, whether we should check the all-fit-in-memory case for rocksdb is a question. What's your opinion? Thanks.

carp84:
Here comes the total time and stability of the results from 2 rounds:

* JMH configuration:
```
JMH version: 1.19
VM version: JDK 1.8.0_102, VM 25.102-b52
VM options: -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.ssl
Warmup: 10 iterations, 1 s each
Measurement: 10 iterations, 1 s each
Timeout: 10 min per iteration
Threads: 1 thread, will synchronize iterations
Benchmark mode: Throughput, ops/time
```

* Round 1 time: `Run complete. Total time: 00:43:34`

* Round 2 time: `Run complete. Total time: 00:43:44`

* Result stability

Benchmark|Mode|Cnt|Score|Error|Units
---- | --- | --- | --- | --- | ---
HeapListStateBenchmark.test1Update#1|thrpt|30|2439.935|±115.324|ops/ms
HeapListStateBenchmark.test1Update#2|thrpt|30|2415.013|±122.071|ops/ms
HeapListStateBenchmark.test2Add#1|thrpt|30|3140.404|±152.354|ops/ms
HeapListStateBenchmark.test2Add#2|thrpt|30|3051.865|±244.304|ops/ms
HeapListStateBenchmark.test3Get#1|thrpt|30|1940.100|±71.786|ops/ms
HeapListStateBenchmark.test3Get#2|thrpt|30|1869.135|±109.670|ops/ms
HeapListStateBenchmark.test4GetAndIterate#1|thrpt|30|1857.606|±79.844|ops/ms
HeapListStateBenchmark.test4GetAndIterate#2|thrpt|30|1788.352|±93.340|ops/ms
HeapListStateBenchmark.test5AddAll#1|thrpt|30|294.972|±122.322|ops/ms
HeapListStateBenchmark.test5AddAll#2|thrpt|30|285.633|±113.834|ops/ms
HeapMapStateBenchmark.test1Add#1|thrpt|30|1991.866|±109.012|ops/ms
HeapMapStateBenchmark.test1Add#2|thrpt|30|2033.461|±48.315|ops/ms
HeapMapStateBenchmark.test1Update#1|thrpt|30|1409.441|±80.476|ops/ms
HeapMapStateBenchmark.test1Update#2|thrpt|30|1527.825|±53.739|ops/ms
HeapMapStateBenchmark.test2Get#1|thrpt|30|1396.715|±42.510|ops/ms
HeapMapStateBenchmark.test2Get#2|thrpt|30|1413.608|±73.645|ops/ms
HeapMapStateBenchmark.test3Contains#1|thrpt|30|1507.149|±54.452|ops/ms
HeapMapStateBenchmark.test3Contains#2|thrpt|30|1550.425|±76.011|ops/ms
HeapMapStateBenchmark.test4Keys#1|thrpt|30|9311.358|±236.535|ops/ms
HeapMapStateBenchmark.test4Keys#2|thrpt|30|9593.196|±369.914|ops/ms
HeapMapStateBenchmark.test5Values#1|thrpt|30|9100.976|±267.255|ops/ms
HeapMapStateBenchmark.test5Values#2|thrpt|30|9288.619|±209.923|ops/ms
HeapMapStateBenchmark.test6Entries#1|thrpt|30|8723.019|±186.056|ops/ms
HeapMapStateBenchmark.test6Entries#2|thrpt|30|9448.083|±273.444|ops/ms
HeapMapStateBenchmark.test7Iterator#1|thrpt|30|9522.637|±344.523|ops/ms
HeapMapStateBenchmark.test7Iterator#2|thrpt|30|9351.307|±210.986|ops/ms
HeapMapStateBenchmark.test8Remove#1|thrpt|30|1743.482|±74.815|ops/ms
HeapMapStateBenchmark.test8Remove#2|thrpt|30|1903.833|±105.975|ops/ms
HeapMapStateBenchmark.test9PutAll#1|thrpt|30|756.448|±26.231|ops/ms
HeapMapStateBenchmark.test9PutAll#2|thrpt|30|786.871|±38.946|ops/ms
HeapValueStateBenchmark.test1Update#1|thrpt|30|2048.441|±101.184|ops/ms
HeapValueStateBenchmark.test1Update#2|thrpt|30|2022.074|±178.275|ops/ms
HeapValueStateBenchmark.test2Add#1|thrpt|30|5704.857|±442.342|ops/ms
HeapValueStateBenchmark.test2Add#2|thrpt|30|5264.967|±938.853|ops/ms
HeapValueStateBenchmark.test3Get#1|thrpt|30|2001.197|±78.446|ops/ms
HeapValueStateBenchmark.test3Get#2|thrpt|30|2099.319|±96.104|ops/ms
RocksDBListStateBenchmark.test1Update#1|thrpt|30|187.203|±11.839|ops/ms
RocksDBListStateBenchmark.test1Update#2|thrpt|30|189.827|±11.974|ops/ms
RocksDBListStateBenchmark.test2Add#1|thrpt|30|191.165|±10.647|ops/ms
RocksDBListStateBenchmark.test2Add#2|thrpt|30|193.191|±9.661|ops/ms
RocksDBListStateBenchmark.test3Get#1|thrpt|30|428.343|±40.097|ops/ms
RocksDBListStateBenchmark.test3Get#2|thrpt|30|403.014|±47.031|ops/ms
RocksDBListStateBenchmark.test4GetAndIterate#1|thrpt|30|421.022|±36.899|ops/ms
RocksDBListStateBenchmark.test4GetAndIterate#2|thrpt|30|433.174|±40.616|ops/ms
RocksDBListStateBenchmark.test5AddAll#1|thrpt|30|102.472|±55.705|ops/ms
RocksDBListStateBenchmark.test5AddAll#2|thrpt|30|106.704|±54.815|ops/ms
RocksDBMapStateBenchmark.test1Add#1|thrpt|30|349.681|±36.756|ops/ms
RocksDBMapStateBenchmark.test1Add#2|thrpt|30|342.682|±40.899|ops/ms
RocksDBMapStateBenchmark.test1Update#1|thrpt|30|350.764|±31.309|ops/ms
RocksDBMapStateBenchmark.test1Update#2|thrpt|30|354.110|±37.487|ops/ms
RocksDBMapStateBenchmark.test2Get#1|thrpt|30|45.117|±0.729|ops/ms
RocksDBMapStateBenchmark.test2Get#2|thrpt|30|45.715|±0.820|ops/ms
RocksDBMapStateBenchmark.test3Contains#1|thrpt|30|45.824|±0.620|ops/ms
RocksDBMapStateBenchmark.test3Contains#2|thrpt|30|46.671|±0.592|ops/ms
RocksDBMapStateBenchmark.test4Keys#1|thrpt|30|323.453|±12.886|ops/ms
RocksDBMapStateBenchmark.test4Keys#2|thrpt|30|318.254|±10.652|ops/ms
RocksDBMapStateBenchmark.test5Values#1|thrpt|30|320.683|±9.605|ops/ms
RocksDBMapStateBenchmark.test5Values#2|thrpt|30|316.464|±13.635|ops/ms
RocksDBMapStateBenchmark.test6Entries#1|thrpt|30|239.602|±11.713|ops/ms
RocksDBMapStateBenchmark.test6Entries#2|thrpt|30|240.108|±13.358|ops/ms
RocksDBMapStateBenchmark.test7Iterator#1|thrpt|30|322.243|±9.507|ops/ms
RocksDBMapStateBenchmark.test7Iterator#2|thrpt|30|316.570|±11.002|ops/ms
RocksDBMapStateBenchmark.test8Remove#1|thrpt|30|356.717|±25.874|ops/ms
RocksDBMapStateBenchmark.test8Remove#2|thrpt|30|351.623|±29.733|ops/ms
RocksDBMapStateBenchmark.test9PutAll#1|thrpt|30|88.432|±6.160|ops/ms
RocksDBMapStateBenchmark.test9PutAll#2|thrpt|30|87.454|±5.779|ops/ms
RocksDBValueStateBenchmark.test1Update#1|thrpt|30|343.421|±23.262|ops/ms
RocksDBValueStateBenchmark.test1Update#2|thrpt|30|337.099|±23.011|ops/ms
RocksDBValueStateBenchmark.test2Add#1|thrpt|30|351.838|±19.350|ops/ms
RocksDBValueStateBenchmark.test2Add#2|thrpt|30|330.423|±27.198|ops/ms
RocksDBValueStateBenchmark.test3Get#1|thrpt|30|556.239|±25.828|ops/ms

pnowojski:
@carp84 that's a lot of benchmarks :) 

What is going on with `HeapValueStateBenchmark.test2Add#2`? (It had a huge spread)

I'm thinking about modifying our speed center setup, so that it won't be flooded/overloaded with number of benchmarks. I think a solution might be to start using `projects`. We could keep current benchmarks in `Flink` project, while we could add all of the benchmarks from this PR to a `State Backends` projects. I have played along with this by manually adding some results and you can see [the result here](http://codespeed.dak8s.net:8000/changes/) on the left:
```
Executable
  Flink // <----- project #1
    - Flink 
  State Backends // <------ project #2
    - State Backends
```
It looks like this would allow us to group the benchmark & results together.

In order to make it fully work, we would need couple of more things:

1. modify `save_jmh_result.py` script and add optional parameters for `project` and `executable` (lines 61 & 62).
2. research how could we differentiate between various type of benchmarks here. Currently there is a jenkins job that runs all benchmarks defined in this repository and then uploads them using those two commands:
```
sh "mvn -Dflink.version=`cat ../flink-version` clean install exec:exec"
sh 'python save_jmh_result.py --environment Hetzner --branch master --commit COMMIT --codespeed URL'
```
we would need to research how can we modify the first command to execute either `Flink` or `State Backends` benchmarks. Then we would upload them using second command (two different executions) by passing correct values for the parameters added in step 1.

Thanks to that we might be able to have two independent jenkins jobs - one for `State Backends` and another for `Flink` benchmarks. Could come in handy if you do not want to wait for all of the benchmarks to complete and you are only interested in one of them.

Do you think it makes sense?

carp84:
> What is going on with HeapValueStateBenchmark.test2Add#2?

Didn't notice this and probably due to environment variance. Let me double check.

> We could keep current benchmarks in `Flink` project, while we could add all of the benchmarks from this PR to a `State Backends` projects.

Agreed to use separate projects. Only that the naming seems a little bit strange since backends also belong to Flink (smile). And internally we also have micro benchmarks for checkpoint and timer-service and plan to upstream later (as mentioned above), so maybe it worth some efforts on categorizing.

> we would need to research how can we modify the first command to execute either Flink or State Backends benchmarks.

Please check the [3rd commit](https://github.com/dataArtisans/flink-benchmarks/pull/13/commits/93c5505c3cc4904a06c73ce095599f38057eb780) here, which generate a shaded jmh jar and we could use command like `java -jar target/benchmarks.jar -rf csv org.apache.flink.state.benchmark.*` to run different benchmark (and save result separately) against the shaded jar w/ the commit, which is the way how I generated the above results.

pnowojski:
> Only that the naming seems a little bit strange since backends also belong to Flink (smile).

Agree :( However we are re-using here a [code speed tool](https://github.com/tobami/codespeed) from PyPy project, that is not widely adopted and seems tailor suited just for their single use case, so unless we want to develop UI from scratch or modify code speed, we have to dance around those kind of issues 😒 I'm open to other suggestions.

> `java -jar target/benchmarks.jar -rf csv org.apache.flink.state.benchmark.*`

This looks ok as long as we will be able to integrate this command with jenkins job running the benchmarks.

carp84:
Have setup a [demo](http://codespeed.dak8s.net:8000/timeline/#/?exe=3&ben=grid&env=2&revs=200&equid=off&quarts=on&extr=on) in our speed center and it successfully reflected the effect of a recent improvement:
image

image

carp84:
@StefanRRichter @pnowojski Mind take a look at the latest commit and let me know if any comments? Thanks.

And if current codes look good, I plan to remove the numbering in method names (like from `test1ListUpdate` to `testListUpdate`) and clean up all demo data on our codespeed center.

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmarkBase.java
```diff
@@ -79,19 +87,37 @@ public void tearDown() {
 
     @Benchmark
     public void test1Update() throws Exception {
-        keyedStateBackend.setCurrentKey(random.nextLong(setupKeyCount));
-        listState.update(Collections.singletonList(random.nextLong()));
+        int currentIndex = getCurrentIndex();
```
pnowojski: There is actually a better way to this test initialisation, please check this:

http://tutorials.jenkov.com/java-performance/jmh.html#state-setup-and-teardown

the idea is that you should do the time consuming setup in in context, once per `Level.Trial` (the default), `Level.Iteration` or `Level.Invocation`.

Secondary concern is when you are passing input parameters not through the context, the code is subject to constant foldings/dead code elimination etc: http://tutorials.jenkov.com/java-performance/jmh.html#writing-good-benchmarks

To avoid those kind of issues, a good benchmark should get it's input data through JMH's `@State` and should return something from the benchmark methods. (`void` in `public void test1Update` is also dangerous)

(the same applies to other benchmarks)

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmarkBase.java
```diff
@@ -79,19 +87,37 @@ public void tearDown() {
...
+        keyedStateBackend.setCurrentKey(setupKeys.get(currentIndex % setupKeyCount));
+        listState.update(Collections.singletonList(randomValues.get(currentIndex % randomValueCount)));
```
pnowojski: here you are also benchmarking the creation of the list, maybe consider creating the list in `@Setup` context and benchmarking only `setCurrentKey()` and `update()`?

(the same applies to other benchmarks)

pnowojski (changes requested):
Thanks @carp84 for the update and integration with Jenkins :) It looks good. Couple of comments from my side.

Also I would actually also drop `test` prefix from the test names (rename `test1ListUpdate` to just `listUpdate`). I know that this doesn't  follow the usual (and ours) Java coding style convention, but in the UI this `test` prefix doesn't help with anything and just takes more place.

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmarkBase.java
carp84: Thanks for the reference of `State` object setUp/tearDown annotation, will init the inputs of benchmark methods following this way.

Also checked the reference to prevent constant folding/dead code and confirmed current codes are in good shape (already using `BlackHole` or having `return` statement for getter benchmark, and all benchmark with no return value have specific operations such as `HeapListState.update`)

pnowojski (approved):
I think the change LGTM. Unfortunately we do not have any CI hooked in here, so I assume everything compiles and works well with Jenkins? :)

@StefanRRichter do you have some more comments? (you have pending "changes requested").

StefanRRichter:
@pnowojski No, my request have been addressed.

StefanRRichter (approved):

carp84:
> so I assume everything compiles and works well with Jenkins?

Yes, please refer to this [Jenkins job](http://codespeed.dak8s.net:8080/job/statebackend-benchmark-test/28/) which is a dry run. :-)

And thanks all for review! @pnowojski @StefanRRichter

pnowojski:
Ok :)

One last comment. Can you @carp84 squash the commits together, except of `Add support to generate shaded benchmark package to allow run specifi...` and can you update/modify the README.md to document the new way to run only selected benchmarks? (basically the `java -jar target/benchmarks.jar -rf csv org.apache.flink.state.benchmark.*` command)

carp84:
Updated, please check and let me know whether it looks good, thanks. @pnowojski

pnowojski (approved):
LGTM, merging. Thanks for the big contribution @carp84 !

PR #14, opened 2019-05-01
[hotfix] Missing quotes in example maven command in README

carp84:
This is a simple fix to add missing quotes in README.


carp84:
@pnowojski My bad to miss the quotes in example command when modifying README and here is a hot fix. Thanks.

pnowojski (approved):
Np :)

PR #15, opened 2019-05-07
[hotfix] Exclude state benchmark from the Flink suit

carp84:
Exclude state benchmark from Flink suit to prevent duplicate testing.


carp84:
Test is running on [Jenkins](http://codespeed.dak8s.net:8080/job/test-exclude-state-benchmark/1/)

pnowojski (approved):

PR #16, opened 2019-05-09
[hotfix] Fix NoClassDefFoundError when running network related benchmarks

carp84:
Network benchmarks are no longer executed since 29th of April and checking the [Jenkins log](http://codespeed.dak8s.net:8080/job/flink-master-benchmarks/3699/console) we could see below error:

> java.lang.NoClassDefFoundError: org/mockito/stubbing/Answer
> 	at org.apache.flink.streaming.runtime.io.benchmark.StreamNetworkBenchmarkEnvironment.createInputGate(StreamNetworkBenchmarkEnvironment.java:262)

This is caused by FLINK-12199 where `StreamNetworkBenchmarkEnvironment` is modified to invoke methods of `InputChannelTestUtils` thus involving `org.mockito.stubbing.Answer`.

Since `mockito-core` is a test-scope dependency of `flink-runtime` thus not a transitive dependency to `flink-benchmarks`, we need to explicitly add `mockito-core` as a dependency in `flink-benchmarks` project to fix the issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,7 @@ under the License.
 		<jmh.version>1.19</jmh.version>
 		<junit.version>4.12</junit.version>
 		<avro.version>1.8.2</avro.version>
+		<mockito.version>2.21.0</mockito.version>
 	</properties>
 
 	<repositories>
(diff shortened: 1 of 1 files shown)


carp84:
Launched a [test job](http://codespeed.dak8s.net:8080/job/hotfix-test/1/) on Jenkins for verification.

carp84:
Confirmed patch could work [perfectly](http://codespeed.dak8s.net:8080/job/hotfix-test/1/artifact/jmh-result.csv/*view*/).

pnowojski (approved):
Thanks for finding the solution to this issue. This is really annoying that Maven works this way.

PR #17, opened 2019-05-10
Run all benchmarks once on Travis CI

pnowojski:
This PR defines mvn test command and runs it on Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: java
+
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,42 @@ under the License.
 			</plugin>
 
 			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>exec-maven-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>test-benchmarks</id>
+						<phase>test</phase>
+						<goals>
+							<goal>exec</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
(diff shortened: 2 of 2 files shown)


pnowojski:
One problem here is that this uses flink version as specified in the pom.xml, while jenkins benchmark runs they use flink version from the flink/master branch. As long as we keep this `pom.xml` up to date, they will be running the same code.

carp84 (approved):
+1, LGTM. Thanks.

NicoK:
@pnowojski can you share a rational as to why to add this? Is it a form of testing whether the benchmarks run / to verify PRs? In that case, it does not really matter anyway whether the executed version is different or not.

pnowojski:
Yes, it's for a verification purpose. Unfortunately @NicoK it matters, take a look at this issue https://github.com/dataArtisans/flink-benchmarks/pull/16 . Change in Flink repository caused network benchmarks to fail because of some dependency issues. Also someone can brake the compatibility in the classes that `flink-benchmarks` are using.

NicoK:
But then simply adding Travis runs after commits to `flink-benchmarks` wouldn't help (for this #16). You'd need daily travis runs like in Flink and notify someone of this failure. Also, you could get the latest revision of the Flink repo and run it just like in our test environment.

pnowojski:
I think I configured somewhere that travis build will be scheduled once couple of days or once a day if it was inactive. True, we could use of some notifications. I wasn't claiming that this PR will solve all the problems and bring peace to the world :) It cheaply makes things a bit better.

PR #18, opened 2019-05-16
Use maven profile to differentiate travis and jenkins run

carp84:
Change of https://github.com/dataArtisans/flink-benchmarks/pull/17 added a new `execution` and attached it to the `test` phase. In our Jenkins configuration we don't add `-DskipTests` option thus will also run the newly added execution, which will mess up the benchmark result.

Simply adding `-DskipTests` to Jenkins configuration and making the `test-benchmark` execution aware of it could work, but this will also [require](https://stackoverflow.com/questions/2192660/maven-maven-exec-plugin-multiple-execution-configurations/31338366#31338366) maven version to be above 3.3.1 and using `mvn exec:exec@run-benchmark` for master run, which seems a little bit hack. So here we suggest to use the `profile` way to differentiate `travis` and `jenkins` run

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@ under the License.
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<flink.version>1.8-SNAPSHOT</flink.version>
+		<flink.version>1.9-SNAPSHOT</flink.version>
 		<java.version>1.8</java.version>
 		<scala.binary.version>2.11</scala.binary.version>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
--- a/pom.xml
+++ b/pom.xml
@@ -154,7 +154,7 @@ under the License.
 
 	<profiles>
 		<profile>
-			<id>travis</id>
+			<id>test</id>
 			<activation>
 				<activeByDefault>false</activeByDefault>
 			</activation>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -150,129 +152,179 @@ under the License.
...
-						<argument>org.apache.flink.benchmark.full.*,org.apache.flink.state.benchmark.*</argument>
-						<argument>-rf</argument>
-						<argument>csv</argument>
-						<argument>.*</argument>
-					</arguments>
-				</configuration>
-			</plugin>
+	<profiles>
+		<profile>
+			<id>travis</id>
```
pnowojski: nit: I would rename travis -> tests & jenkins -> benchmarks, to name them after function they do instead of where are they executed.

pom.xml
```diff
@@ -150,129 +152,179 @@ under the License.
...
+								<argument>-wi</argument>
+								<argument>0</argument>
+								<argument>-rf</argument>
+								<argument>csv</argument>
+								<argument>.*</argument>
+							</arguments>
+						</configuration>
+					</plugin>
+
+					<plugin>
```
pnowojski: Could we deduplicate/share `maven-compiler-plugin` and `avro-maven-plugin` across both profiles?

pnowojski (changes requested):
Thank you @carp84  for the fix! I've left two smaller comments. 

Could you update the README.md and document that in order to run tests one should execute following command:
```
mvn test -P travis
// or
mvn test -P test
```

pnowojski (approved):
LGTM, thanks for the fix

pnowojski:
I have squashed some commits and merged it manually.

PR #20, opened 2019-05-20
Add InputBenchmark

StefanRRichter:
This PR adds additional benchmarks for performance of inputs.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/InputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/InputBenchmark.java
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -20,10 +20,10 @@ package org.apache.flink.benchmark;
 
 import org.apache.flink.benchmark.functions.LongSource;
 import org.apache.flink.benchmark.functions.MultiplyByTwo;
-import org.apache.flink.benchmark.functions.SumReduce;
 import org.apache.flink.streaming.api.datastream.DataStreamSource;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
-import org.apache.flink.streaming.api.windowing.assigners.GlobalWindows;
+import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.OperationsPerInvocation;
 import org.openjdk.jmh.runner.Runner;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/flink/benchmark/InputBenchmark.java
```diff
@@ -0,0 +1,78 @@
...
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+@OperationsPerInvocation(value = InputBenchmark.RECORDS_PER_INVOCATION)
+public class InputBenchmark {
```
pnowojski: Can you add `extends BenchmarkBase`? It contains some common settings that we want to be shared among benchmarks.

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
```diff
@@ -0,0 +1,78 @@
...
+		Options options = new OptionsBuilder()
+			.verbosity(VerboseMode.NORMAL)
+			.include(".*" + InputBenchmark.class.getSimpleName() + ".*")
+			.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void benchmarkChained(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: Regarding naming the benchmark methods, there is one important thing. Results uploader is using just the benchmark method name + parameters to generate benchmark name visible in the [codespeed](http://codespeed.dak8s.net:8000). Because of that it is important to:
1. have the method name explanatory enough so we can quickly remember what the benchmark is all about (
2. be as short as possible to not bloat the web UI (so we must drop all of the redundant suffixes/prefixes) 

I guess you were basing on `benchmarkCount`, but it is actually a terrible name, which was kept because of historical reasons. More recent benchmarks are named in a much better way.

Maybe rename `benchmarkChained` -> `chainedMap` and `benchmarkRebalance` -> `rebalancedMap`? Or `mapSink` & `mapRebalanceMapSink`?

pnowojski: If we are adding this, I think we should drop the `benchmarkCount` benchmark.

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
```diff
@@ -0,0 +1,78 @@
...
+			.map(new MultiplyByTwo())
+			.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	public void benchmarkChannel(FlinkEnvironmentContext context) throws Exception {
+
+		StreamExecutionEnvironment env = context.env;
```
pnowojski, line 69: enable checkpointing in both benchmarks? The tricky part will be to find a good checkpointing interval value, since those benchmarks are executed within 500ms-1000ms.  

If we drop the `benchmarkCount`, we would loose non checkpointed coverage and in that case the question is does it matter for us? If after enabling checkpointing interval the throughput doesn't change by much, it probably doesn't matter.

pnowojski (changes requested):
Thanks for the new benchmarks @StefanRRichter. I've left couple of comments.

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
StefanRRichter: Oh yes, I was planning to do that but seems I simply forgot it :D

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
StefanRRichter: As for the naming, I think it depends on what we want to describe. Do we care more about the details of what it does (map, rebalance...) or more what we want to control with it (general pipeline speed for simple chained jobs or jobs that go through a channel). I am happy if we can come up with better names, I never liked those very much. Let's just quickly agree about what we want to express with those names.

StefanRRichter: No objections, think it does not offer much additional value.

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
StefanRRichter: Just to be clear, would you like to enable it in general, or have it somehow parameterized/separate benchmark case? In my head, that almosts tests two different aspects of the system.

StefanRRichter:
Updated.

src/main/java/org/apache/flink/benchmark/InputBenchmark.java
pnowojski: 100ms checkpointing interval should be fine. We can adjust it in the future if we see the need.

pnowojski (approved):
LGTM, thanks for the update!

StefanRRichter:
Thanks for review & merge!

PR #21, opened 2019-05-28
Add TwoInputBenchmark

sunhaibotb:
Adds a high level  benchmark based on DataStream API for TwoInputStreamOperator.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.benchmark.functions.LongSource;
+import org.apache.flink.benchmark.operators.MultiplyByTwoCoStreamMap;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+
+import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	@Benchmark
+	public void mapSink(FlinkEnvironmentContext context) throws Exception {
+
+		StreamExecutionEnvironment env = context.env;
+		env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
+		env.setParallelism(1);
+
+		DataStreamSource<Long> source1 = env.addSource(new LongSource(RECORDS_PER_INVOCATION));
```
pnowojski: `RECORDS_PER_INVOCATION / 2` to have the total number or records inlined? However I'm not sure what should be our convention here.

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
```diff
@@ -0,0 +1,68 @@
...
+		Options options = new OptionsBuilder()
+			.verbosity(VerboseMode.NORMAL)
+			.include(".*" + TwoInputBenchmark.class.getSimpleName() + ".*")
+			.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void mapSink(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: This name is colliding with already existing benchmark. I've added an explanation [to the README](https://github.com/dataArtisans/flink-benchmarks/#naming-convention) about the naming convention/restrictions that we have, please take a look at it.

Maybe `twoInputMapSink`?

pnowojski (changes requested):
Thanks @sunhaibotb. I've left two smaller comments.

After resolving the comments, could you provide the output of running the new benchmark with results and how much did the benchmarking take time?

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
sunhaibotb: > RECORDS_PER_INVOCATION / 2 to have the total number or records inlined?

You are right.

sunhaibotb:
I have update the code.  The benchmarking takes `02:22 minutes` when `RECORDS_PER_INVOCATION` is `15,000,000`,  and the output of benchmarking is as follow. @pnowojski 

```
Result "org.apache.flink.benchmark.TwoInputBenchmark.twoInputMapSink":
  8014.500 ±(99.9%) 465.530 ops/ms [Average]
  (min, avg, max) = (6150.633, 8014.500, 9516.431), stdev = 696.783
  CI (99.9%): [7548.970, 8480.029] (assumes normal distribution)


# Run complete. Total time: 00:02:08

Benchmark                           Mode  Cnt     Score     Error   Units
TwoInputBenchmark.twoInputMapSink  thrpt   30  8014.500 ± 465.530  ops/ms
```

sunhaibotb:
I changed `RECORDS_PER_INVOCATION` from `15,000,000` to `25,000,000`, and found that the throughput increased significantly (which means `15,000,000` is not enough). Then increased it to `30,000,000`, the throughput is essentially no longer increasing. Therefore, I finally decided to change `RECORDS_PER INVOCATION` to `25,000,000`, and the code has been update. The benchmarking takes about `3 minutes` when `RECORDS_PER_INVOCATION` is `25,000,000`.

pnowojski (approved):
LGTM

PR #22, opened 2019-06-06
[FLINK-11877] Add an input idle benchmark for two input operators

sunhaibotb:
Adds a high level benchmark based on DataStream API for two-input operators to track a potential regression when one input is idle.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
@@ -20,6 +20,7 @@ package org.apache.flink.benchmark;
 
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.benchmark.functions.LongSource;
+import org.apache.flink.benchmark.functions.QueuingLongSource;
 import org.apache.flink.benchmark.operators.MultiplyByTwoCoStreamMap;
 import org.apache.flink.streaming.api.datastream.DataStreamSource;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
@@ -33,10 +34,12 @@ import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 import org.openjdk.jmh.runner.options.VerboseMode;
 
-@OperationsPerInvocation(value = TwoInputBenchmark.RECORDS_PER_INVOCATION)
 public class TwoInputBenchmark extends BenchmarkBase {
 
 	public static final int RECORDS_PER_INVOCATION = 25_000_000;
+
+	public static final int ONEIDLE_RECORDS_PER_INVOCATION = 15_000_000;
+
 	private static final long CHECKPOINT_INTERVAL_MS = 100;
 
 	public static void main(String[] args)
@@ -50,6 +53,7 @@ public class TwoInputBenchmark extends BenchmarkBase {
 	}
 
 	@Benchmark
+	@OperationsPerInvocation(value = TwoInputBenchmark.RECORDS_PER_INVOCATION)
 	public void twoInputMapSink(FlinkEnvironmentContext context) throws Exception {
 
 		StreamExecutionEnvironment env = context.env;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
```diff
@@ -67,4 +69,42 @@ public void twoInputMapSink(FlinkEnvironmentContext context) throws Exception {
 
 		env.execute();
 	}
+
+	@Benchmark
+	public void twoInputSelectableMapSink(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: I think I wouldn't be adding this benchmark. The plan is to quickly drop non selectable `TwoInputProcessor` and in that case, `twoInputSelectableMapSink` would duplicate `twoInputMapSink`. In order to save some benchmarking time and not clog the UI, I would keep only `twoInputMapSink`.

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
```diff
@@ -67,4 +69,42 @@ public void twoInputMapSink(FlinkEnvironmentContext context) throws Exception {
...
+		source1
+				.connect(source2)
+				.transform("custom operator", TypeInformation.of(Long.class), new MultiplyByTwoAndInputSelectableCoStreamMap())
+				.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	public void twoInputSelectableAndOneAvailableMapSink(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: Kind of the same argument applies to this one. 

I would rewrite this `twoInputSelectableAndOneAvailableMapSink` benchmark, to non selectable version (keeping `QueuingLongSource`). My reasoning is that it would be more valuable to track a potential regression once we drop old `StreamTwoInputProcessor`.

Alternate solution would be to have both selectable and second one non selectable for this scenario, but it would again I think unnecessarily increase the number of benchmarks.

If you agree with this reasoning, I would just use `MultiplyByTwoCoStreamMap` here and rename `twoInputSelectableAndOneAvailableMapSink` -> to `twoInputOneIdleMapSink`?

pnowojski (changes requested):
Thanks for the contribution @sunhaibotb. I've left couple of comments.

Could you also mark the commit with `[FLINK-11877]` prefix? As the benchmarks are now officially a Flink component( `[DISCUSS] Clean up and reorganize the JIRA components` dev mailing list), I think we could start marking/linking them to Flink's issues.

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
sunhaibotb: The plan gives good consideration to the near future. I agree.

sunhaibotb:
The PR has been updated. @pnowojski

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
```diff
@@ -67,4 +68,23 @@ public void twoInputMapSink(FlinkEnvironmentContext context) throws Exception {
...
+
+	@Benchmark
+	public void twoInputOneIdleMapSink(FlinkEnvironmentContext context) throws Exception {
+
+		StreamExecutionEnvironment env = context.env;
+		env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
+		env.setParallelism(1);
+
+		QueuingLongSource.reset();
+		DataStreamSource<Long> source1 = env.addSource(new QueuingLongSource(1, RECORDS_PER_INVOCATION - 1));
```
pnowojski: One last thing. I've checked the benchmark results and unfortunately this one is very slow with 25M of records. However it is I think also more stable compared to the non idle one, so I think we should be fine with changing number of records to `15M` for `twoInputOneIdleMapSink`. 

I think that we can put `@OperationsPerInvocation` annotation at the method level to have two benchmarks in the same file running with different number of records.

sunhaibotb: Yes, I'll revise it right away.

pnowojski (approved):
LGTM, thanks for the contribution @sunhaibotb

src/main/java/org/apache/flink/benchmark/TwoInputBenchmark.java
sunhaibotb: The code has been updated. @pnowojski

sunhaibotb:
Thanks for reviewing and merging @pnowojski

PR #23, opened 2019-06-12
[FLINK-12818] Improve stability of twoInputMapSink benchmark

pnowojski:
This is an attempt to improve twoInputMapSink benchmark stability. Local test show that for some unknown reason, without 1ms buffer timeout, some JVM forks are much slower then others, making results unstable.


StefanRRichter (approved):
LGTM 👍

pnowojski:
Thanks @StefanRRichter for quick review. Let's hope that this will fix the issue - we will see tomorrow or in couple of days.

Issue #25, opened 2019-06-27
Delete the tmp directory after benchmark teardown


klion26:
Currently, the statebanchmark did not delete the temp directory after the benchmark teardown, so we'll have many empty directory in `/tmp`,

pnowojski:
Thanks for reporting the issue. 

Since `flink-benchmarks` are now officially a component of Apache Flink project, next time you could use Flink's Apache Jira to fill the issue.

pnowojski:
Issue was resolved by #26

klion26:
thanks for the review and merge, will create Apache Jiar next time~

PR #26, opened 2019-06-28
Delete temp root directory(empty) when teardown the state benchmark

klion26:
As #25 said, currently the state benchmark will not delete the tmp root directory after teardown. so in the directories in `/tmp` will increase forever, this pr wants to fix it.


pnowojski (approved):
Thanks for the fix. LGTM

PR #30, opened 2019-10-02
Adding benchmark for AsyncWaitOperator

AHeise:


src/main/java/org/apache/flink/benchmark/AsyncWaitOperatorBenchmark.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	private static class BenchmarkAsyncFunctionExecutor extends RichAsyncFunction<Long, Long> {
+		private transient ExecutorService executor;
+
+		@Override
+		public void open(Configuration parameters) throws Exception {
+			super.open(parameters);
+
+			executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
pnowojski: Instead of starting up the thread pool and shutting it down once per iteration in the `open()`, we could:
- move this to `setUp()` and `teraDown()` methods per iteration, but wouldn't be measured
- move this to `setUp()` and `teraDown()` methods per JVM instance (I'm not sure but I would guess JMH supports that. If not, we could make it just a static variable).

That would reduce the startup overhead by a bit, however I don't know if in any measurable way?

pnowojski (commented):
Thanks for the contribution. Left one question/comment.

src/main/java/org/apache/flink/benchmark/AsyncWaitOperatorBenchmark.java
AHeise: 👍

pnowojski (approved):
LGTM

Will merge with green CI

PR #31, opened 2019-10-03
Add throughput benchmark executor for data skew scenario.

wsry:
The purpose of this pr is to add a throughput benchmark executor for data skew scenario. The benchmark results are as follows.
```
Benchmark                                                            Mode  Cnt     Score      Error   Units
DataSkewStreamNetworkThroughputBenchmarkExecutor.networkThroughput  thrpt   30  6048.341 ± 1423.567  ops/ms
```

### DIFF
--- a/src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
+++ b/src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.streaming.runtime.io.benchmark.DataSkewStreamNetworkThroughputBenchmark;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import static org.openjdk.jmh.annotations.Scope.Thread;
+
+/**
+ * JMH throughput benchmark runner for data skew scenario.
+ */
+@OperationsPerInvocation(value = DataSkewStreamNetworkThroughputBenchmarkExecutor.RECORDS_PER_INVOCATION)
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+		Options options = new OptionsBuilder()
+				.verbosity(VerboseMode.NORMAL)
+				.include(".*" + DataSkewStreamNetworkThroughputBenchmarkExecutor.class.getSimpleName() + ".*")
+				.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void networkThroughput(MultiEnvironment context) throws Exception {
```
pnowojski: As explained in the [readme](https://github.com/dataArtisans/flink-benchmarks):
> Method names should be unique across the project.

Maybe `networkSkewedThroughput`? (`network` prefix first to group the benchmarks together?)

src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+	@State(Thread)
+	public static class MultiEnvironment extends DataSkewStreamNetworkThroughputBenchmark {
+		// 1ms buffer timeout
+		private final int flushTimeout = 1;
+
+		// 1000 num of channels (subpartitions)
+		private final int channels = 1000;
+
+		// 10 writer threads
+		private final int writers = 10;
```
pnowojski: Why did you pick value `10`? If there is no particular reason, for the sake of uniformity I would stick to the same value as non skewed benchmarks (`4`).

pnowojski (changes requested):
The results are a bit worrying. ~6000 records/ms is pretty small value, but the bigger issue is error of +/- 1400. Can you post the full results of all of the warmup and normal iterations? (Please make sure to not use the computer while benchmarking). 

Also, previously you were posting the results:
Before the fix:
```
Benchmark    (channelsFlushTimeout) Mode     Cnt     Score      Error   Units
NetworkThroughput     1000,1ms      thrpt    30   18240.197 ± 1892.419 ops/ms
```
After the fix:
```
Benchmark    (channelsFlushTimeout) Mode     Cnt     Score      Error  Units
NetworkThroughput     1000,1ms      thrpt    30   24532.313 ± 1118.312 ops/m
```
Where did this sudden change (18k vs 6k) is coming from?

src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
wsry: 10 writers can make the gap between results before and after fix bigger. This is where the sudden change (18k vs 6k) is coming from. The previous result is coming from 4 writers.
The latest results are as follows.
Before fix:
```
Benchmark                                                            Mode  Cnt     Score      Error   Units
DataSkewStreamNetworkThroughputBenchmarkExecutor.networkSkewedThroughput  thrpt   30  6014.192 ± 1090.448  ops/ms
```
After fix:
```
Benchmark                                                            Mode  Cnt      Score     Error   Units
DataSkewStreamNetworkThroughputBenchmarkExecutor.networkSkewedThroughput  thrpt   30  18609.481 ± 715.554  ops/ms
```

src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
wsry: Fixed.

src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
pnowojski: I think 10 writer threads might be a bit unrealistic scenario and can cause excessive thread contention, as the benchmarking machine has only 4 CPU cores. Because of that I would stick with the previous value, as it also might improve benchmark stability.

wsry: Your concern is reasonable. I have changed the number of writers to 4 which also works on my laptop.

pnowojski (approved):
LGTM, merging.

PR #32, opened 2019-10-11
[FLINK-14118] Bump number of writer threads in network data skew Benchmark to 10

pnowojski:
Previous value of 4, was not showing the expected speed improvement on the faster benchmarking machine (compared to slower development laptops).

With this PR, results before the fix:

```
Benchmark                                                                  Mode  Cnt     Score      Error   Units
DataSkewStreamNetworkThroughputBenchmarkExecutor.networkSkewedThroughput  thrpt   15  7300.068 ± 4629.374  ops/ms
```

and after:
```

Benchmark                                                                  Mode  Cnt      Score      Error   Units
DataSkewStreamNetworkThroughputBenchmarkExecutor.networkSkewedThroughput  thrpt   15  20435.596 ± 1092.723  ops/ms
```

While with 4 writer threads, there was no difference on my laptop.

@wsry

### DIFF
--- a/src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
+++ b/src/test/java/org/apache/flink/benchmark/DataSkewStreamNetworkThroughputBenchmarkExecutor.java
@@ -67,8 +67,8 @@ public class DataSkewStreamNetworkThroughputBenchmarkExecutor extends BenchmarkB
 		// 1000 num of channels (subpartitions)
 		private final int channels = 1000;
 
-		// 4 writer threads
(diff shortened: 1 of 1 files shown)


wsry:
@pnowojski You are right, the speed improvement is relevant to the test environment. If the netty thread is fast enough to process all the buffers and flush events. The improvement should be minor, and if more pressure is given to the netty thread, the speed improvement will be more noticeable.

pnowojski:
Thanks for the review. Merging

wsry:
Thanks @pnowojski for making the speed improvement more visible.

PR #33, opened 2019-11-07
[FLINK-13846] Add benchmark of mapstate isEmpty

Myasuka:
Sine we have new API of `mapState#isEmpty()`, we should also have benchmarks on this.

### DIFF
--- a/src/main/java/org/apache/flink/state/benchmark/MapStateBenchmark.java
+++ b/src/main/java/org/apache/flink/state/benchmark/MapStateBenchmark.java
@@ -102,6 +102,12 @@ public class MapStateBenchmark extends StateBenchmarkBase {
     }
 
     @Benchmark
+    public boolean mapIsEmpty(KeyValue keyValue) throws Exception {
+        keyedStateBackend.setCurrentKey(keyValue.setUpKey);
+        return mapState.isEmpty();
+    }
+
+    @Benchmark
     @OperationsPerInvocation(mapKeyCount)
     public void mapKeys(KeyValue keyValue, Blackhole bh) throws Exception {
         keyedStateBackend.setCurrentKey(keyValue.setUpKey);


pnowojski:
As long as benchmark is stable LGTM % the test failure. Can you take a look at it @Myasuka ?

Myasuka:
@pnowojski  This is due to we still use `1.9-SNAPSHOT` as flink version, I changed it to `1.10-SNAPSHOT` now.

Myasuka:
@wsry , the CI failed due to `java.lang.OutOfMemoryError: Could not allocate enough memory segments for NetworkBufferPool` at [CI-report](https://travis-ci.org/dataArtisans/flink-benchmarks/builds/609160507?utm_source=github_status&utm_medium=notification) of `DataSkewStreamNetworkThroughputBenchmarkExecutor`, would you please take a look?

pnowojski:
Upgrade from 1.9-SNAPSHOT to 1.10-SNAPSHOT has caused this failure? What has changed?

Myasuka:
@wsry , as discussed offline previously, we just thought this was an occasional OOM, but my [private CI](https://travis-ci.org/Myasuka/flink-benchmarks/builds/610417111) also failed due to OOM. Please take a look at this problem.

wsry:
I guess the reason is insufficient memory of the test environment. I will open a PR to fix the problem. @Myasuka @pnowojski

wsry:
I found the test VM has more than 7G memory which should be enough. It is the ```exec-maven-plugin``` plugin which does not give enough memory to the test JVM process leading to the problem. Could you bother to fix the problem by adding a hotfix commit to avoid opening a new PR? @Myasuka

pnowojski:
What's the status of this PR @wsry @Myasuka?

wsry:
After discussed offline with @Myasuka , I open a new PR to fix the problem https://github.com/dataArtisans/flink-benchmarks/pull/35. Could you please take a look. @pnowojski

pnowojski:
Thanks @wsry. Could you @Myasuka rebase on the master to let the travis check the PR?

Myasuka:
Thanks the hot fix from @wsry , and already rebased with latest master branch @pnowojski .

pnowojski (approved):
LGTM

PR #34, opened 2019-11-14
[FLINK-14783] Add ContinuousFileReaderOperator benchmark

rkhachatryan:
Duration: the test takes about 2 minutes on my local machine (AsyncWaitOperatorBenchmark takes a little bit more).

jira: https://issues.apache.org/jira/browse/FLINK-14783


src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,133 @@
...
+        absPath = buildPath();
+        fileReader = chooseFormat(absPath.toString(), (fmt.isEmpty()) ? guessFormat(absPath) : fmt);
+
+        Preconditions.checkArgument(absPath.exists(), "%s doesn't exist", absPath);
+        if (absPath.isDirectory()) Preconditions.checkArgument(absPath.canExecute(), "can't list files in %s", absPath);
+        else Preconditions.checkArgument(absPath.canRead(), "can't read %s", absPath);
+    }
+
+    @Benchmark
+    public void readFs(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: maybe rename to `readFile` to be more consistent with the benchmarked method?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,133 @@
...
+@OperationsPerInvocation(value = ContinuousFileReaderOperatorBenchmark.RECORDS_PER_INVOCATION)
+@Fork(value = 1)
+public class ContinuousFileReaderOperatorBenchmark extends BenchmarkBase {
+    public static final int RECORDS_PER_INVOCATION = 1000;
+
+    private static final long CHECKPOINT_INTERVAL_MS = 100;
+
+    private final String BASE_DIR = "src/main/resources/"; // used if non-absolute path provided
+
+    private File absPath;
```
pnowojski: generally speaking we are avoiding abbreviations (abs -> absolute, fmt -> format, dop -> parallelism, ..)

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,133 @@
...
+ * Parameters: <ul>
+ *     <li>path - folder or file; if not absolute, it's assumed relative to src/main/resources </li>
+ *     <li>fmt - format hint (optional: if not given it is inferred from file extension or starting of folder name)</li>
+ *     <li>dop - degree of parallelism.</li>
+ * </ul>
+ */
+// bash script to generate 10 files with 100 random base64 strings each:
+// for f in $(seq 1 10); do for l in $(seq 1 100); do dd if=/dev/random count=2 | base64 >> src/main/resources/txt-10-100-1024/$f; done; done
+@OperationsPerInvocation(value = ContinuousFileReaderOperatorBenchmark.RECORDS_PER_INVOCATION)
+@Fork(value = 1)
```
pnowojski: Please use the default settings defined in the `BenchmarkBase`. `Fork(value = 1)` can introduce instabilities that might not be easily detectable, as different JVM instances can optimise could differently (one fork can be consistently faster while other consistently slower).

To offset the higher number of benchmarking runs, I would think about reducing number of tested parameters combination. Do we need to test with different formats? Or with different parallelism?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,133 @@
...
+public class ContinuousFileReaderOperatorBenchmark extends BenchmarkBase {
+    public static final int RECORDS_PER_INVOCATION = 1000;
```
pnowojski: I think that the number of records here might be too small and the benchmark results might be affected by for example startup overheads, instead of per record overheads that we are trying to measure.

Also the record size is probably not helping here. The larger the records, the less important are overheads of the `ContinuousFileReaderOperator` itself. In order to stress it the most and to test for it's code (not the deserialisation/reading IO/memory copy), the records should be as small as possible and as light weight as possible (binary format > text).

Maybe just `SerializedInputFormat` using a simple `org.apache.flink.types.IntValue`? Keep in mind that we do not want to benchmark here different input formats (avro/csv/text) but the operator itself (input formats could be benchmarked in more "unit test" style, without running a Flink job)

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,133 @@
...
+ * By default it runs with DOP 1 and then 2.
+ * <p>
+ * Parameters: <ul>
+ *     <li>path - folder or file; if not absolute, it's assumed relative to src/main/resources </li>
+ *     <li>fmt - format hint (optional: if not given it is inferred from file extension or starting of folder name)</li>
+ *     <li>dop - degree of parallelism.</li>
+ * </ul>
+ */
+// bash script to generate 10 files with 100 random base64 strings each:
+// for f in $(seq 1 10); do for l in $(seq 1 100); do dd if=/dev/random count=2 | base64 >> src/main/resources/txt-10-100-1024/$f; done; done
```
pnowojski: I'm thinking whether it's better to have static resource file or a file generated in the `setUp` method?

pnowojski (commented):
Thanks for the benchmark. I've left couple of comments.

rkhachatryan:
Addressed most of the issues.

Regarding SerializedInputFormat, I don't think it makes a big difference, but text is just simpler. And same for static files vs generation on the fly.

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+
+        @Override
+        public String nextRecord(String s) {
+            count++;
+            return LINE;
+        }
+
+        @Override
+        public void open(FileInputSplit fileSplit) {
+            // prevent super from accessing file
```
pnowojski, line 109: missing 
```
count = 0;
```
As it is now, the throughput is inflated 100 times 👿 Probably the the `LINES_PER_SPLIT` should be decreased as a side effect to `200_000`?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Benchmark
+    public void readFiles(FlinkEnvironmentContext context) throws Exception {
+        StreamExecutionEnvironment env = context.env;
+        env
+                .enableCheckpointing(100)
+                .setParallelism(1)
+                .addSource(buildSourceFun())
+                .transform("fileReader", TypeInformation.of(String.class),
+                        new ContinuousFileReaderOperator<>(new MockInputFormat()))
```
pnowojski: instead of using `@Internal` classes, can the same logic be tested by using `#readFile` API? Creating some empty temporary files in a temporary directory?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+
+        @Override
+        public boolean reachedEnd() {
+            return count >= ContinuousFileReaderOperatorBenchmark.LINES_PER_SPLIT;
+        }
+
+        @Override
+        public String nextRecord(String s) {
+            count++;
+            return LINE;
```
pnowojski, line 103: nit `return LINE + count;` to make jit's live more difficult?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+            }
+
+            @Override
+            public void cancel() {
+                isRunning = false;
+            }
+        };
+    }
```
pnowojski: convert into a named `private static` class ? Calling `buildSourceFun()` is a bit strange equivalent of calling a constructor.

pnowojski (commented):
Thanks for the update @rkhachatryan. Left a couple of comments :)

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+        Options options = new OptionsBuilder()
+                .verbosity(VerboseMode.NORMAL)
+                .include(".*" + ContinuousFileReaderOperatorBenchmark.class.getSimpleName() + ".*")
+                .build();
+
+        new Runner(options).run();
+    }
+
+    @Benchmark
+    public void readFiles(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: If we still want to remain reference to the `#readFile()` method: `readFiles` -> `readFile` ? If not, `readFiles` -> `fileReader`?

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
rkhachatryan: It requires a lot of knowledge of how it internally works (configuration, file splits by source function). So will also break in case of changes.
Also, it means testing CFRO in less isolation.

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
rkhachatryan: jit shouldn't touch it, as we are writing to a variable here.
With `LINE + count` performance degrades, but insignificantly, meaning we are just concatenating strings now in addition to regular CFRO work.

pnowojski: I wouldn't be so sure - this method call should be devirtualized and inlined. As this returns constant in some loop, magic could happen. JVM is relatively stupid compared to gcc/llvm, but still...

This one is worth reading
http://tutorials.jenkov.com/java-performance/jmh.html#constant-folding

But we can leave it as it is.

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
pnowojski: `readFileSplit` as proposed in the offline discussion?

pnowojski (approved):
% `readFileSplit` LGTM.

Could you also rebase on the latest master (that includes a fix for the failing tests) to get green travis?

rkhachatryan:
Fixed and rebased. 
Thanks for the feedback!

pnowojski:
Thanks! Merged.

PR #36, opened 2019-11-29
[FLINK-14346] [serialization] Add string-heavy version of SerializationFrameworkMiniBenchmark

shuttie:
This PR is a follow-up performance benchmark for the [FLINK-14346](https://issues.apache.org/jira/browse/FLINK-14346) issue about perf improvement for `StringValue.writeString` and `StringValue.readString`. Main PR for the feature is here: https://github.com/apache/flink/pull/10358

Current benchmark results are:
```
Original:

Benchmark                                                  Mode  Cnt    Score    Error   Units
SerializationFrameworkMiniBenchmarks.serializerHeavyPojo  thrpt   30  100.689 ±  1.716  ops/ms
SerializationFrameworkMiniBenchmarks.serializerKryo       thrpt   30  211.890 ±  5.048  ops/ms
SerializationFrameworkMiniBenchmarks.serializerPojo       thrpt   30  473.830 ± 11.037  ops/ms
SerializationFrameworkMiniBenchmarks.serializerRow        thrpt   30  549.780 ± 11.578  ops/ms
SerializationFrameworkMiniBenchmarks.serializerTuple      thrpt   30  587.982 ±  8.941  ops/ms

----------

Improved:

Benchmark                                                  Mode  Cnt    Score   Error   Units
SerializationFrameworkMiniBenchmarks.serializerHeavyPojo  thrpt   30  121.078 ± 2.657  ops/ms
SerializationFrameworkMiniBenchmarks.serializerKryo       thrpt   30  220.956 ± 7.624  ops/ms
SerializationFrameworkMiniBenchmarks.serializerPojo       thrpt   30  445.025 ± 6.894  ops/ms
SerializationFrameworkMiniBenchmarks.serializerRow        thrpt   30  521.074 ± 9.481  ops/ms
SerializationFrameworkMiniBenchmarks.serializerTuple      thrpt   30  545.966 ± 8.708  ops/ms
```

Most of these results are kind of noisy, but the most important one, `serializerHeavyPojo` has +20% performance gain with high confidence.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
+++ b/src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
@@ -25,6 +25,7 @@ import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.tuple.Tuple8;
 import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
 import org.apache.flink.benchmark.functions.BaseSourceWithKeyRange;
+import org.apache.flink.benchmark.full.StringSerializationBenchmark;
 import org.apache.flink.streaming.api.environment.LocalStreamEnvironment;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
@@ -39,6 +40,8 @@ import org.openjdk.jmh.runner.options.OptionsBuilder;
 import org.openjdk.jmh.runner.options.VerboseMode;
 
 import java.util.Arrays;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * Benchmark for serializing POJOs and Tuples with different serialization frameworks.
@@ -66,6 +69,22 @@ public class SerializationFrameworkMiniBenchmarks extends BenchmarkBase {
 		executionConfig.registerPojoType(MyOperation.class);
 
 		env.addSource(new PojoSource(RECORDS_PER_INVOCATION, 10))
+				.rebalance()
+				.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkMiniBenchmarks.RECORDS_PER_INVOCATION)
+	public void serializerHeavyString() throws Exception {
+		LocalStreamEnvironment env =
+				StreamExecutionEnvironment.createLocalEnvironment(1);
+		ExecutionConfig executionConfig = env.getConfig();
+		executionConfig.registerPojoType(MyPojo.class);
+		executionConfig.registerPojoType(MyOperation.class);
+
+		env.addSource(new LongStringSource(RECORDS_PER_INVOCATION, 12))
 				.rebalance()
 				.addSink(new DiscardingSink<>());
 
@@ -126,6 +145,40 @@ public class SerializationFrameworkMiniBenchmarks extends BenchmarkBase {
 				.addSink(new DiscardingSink<>());
 
 		env.execute();
+	}
+
+	/**
+	 * Source emitting a long String.
+	 */
+	public static class LongStringSource extends BaseSourceWithKeyRange<String> {
+		private static final long serialVersionUID = 3746240885982877398L;
+		private String[] templates;
+
+		public LongStringSource(int numEvents, int numKeys) {
+			super(numEvents, numKeys);
+		}
+
+		@Override
+		protected void init() {
+			super.init();
+			templates = new String[] {
+					makeString(StringSerializationBenchmark.asciiChars, 1024),
+					makeString(StringSerializationBenchmark.russianChars, 1024),
+					makeString(StringSerializationBenchmark.chineseChars, 1024)
+			};
+		}
+
+		private String makeString(char[] symbols, int length) {
+			char[] buffer = new char[length];
+			Random random = ThreadLocalRandom.current();
+			Arrays.fill(buffer, symbols[random.nextInt(symbols.length)]);
+			return new String(buffer);
+		}
+
+		@Override
+		protected String getElement(int keyId) {
+			return templates[keyId % templates.length];
+		}
 	}
 
 	/**
--- a/src/main/java/org/apache/flink/benchmark/full/PojoSerializationBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/full/PojoSerializationBenchmark.java
@@ -1,0 +1,127 @@
+package org.apache.flink.benchmark.full;
+import org.apache.flink.api.common.ExecutionConfig;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.java.typeutils.runtime.kryo.KryoSerializer;
+import org.apache.flink.benchmark.BenchmarkBase;
+import org.apache.flink.benchmark.SerializationFrameworkMiniBenchmarks;
+import org.apache.flink.core.memory.DataInputView;
+import org.apache.flink.core.memory.DataInputViewStreamWrapper;
+import org.apache.flink.core.memory.DataOutputView;
+import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
+import org.apache.flink.formats.avro.typeutils.AvroSerializer;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+import org.openjdk.jmh.annotations.*;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@BenchmarkMode({Mode.Throughput})
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class PojoSerializationBenchmark extends BenchmarkBase {
+
+    SerializationFrameworkMiniBenchmarks.MyPojo pojo;
+    org.apache.flink.benchmark.avro.MyPojo avroPojo;
+
+    ExecutionConfig config = new ExecutionConfig();
+    TypeSerializer<SerializationFrameworkMiniBenchmarks.MyPojo> pojoSerializer =
+            TypeInformation.of(SerializationFrameworkMiniBenchmarks.MyPojo.class).createSerializer(config);
+    TypeSerializer<SerializationFrameworkMiniBenchmarks.MyPojo> kryoSerializer =
+            new KryoSerializer<>(SerializationFrameworkMiniBenchmarks.MyPojo.class, config);
+    TypeSerializer<org.apache.flink.benchmark.avro.MyPojo> avroSerializer =
+            new AvroSerializer<>(org.apache.flink.benchmark.avro.MyPojo.class);
+
+    ByteArrayInputStream pojoBuffer;
+    ByteArrayInputStream avroBuffer;
+    ByteArrayInputStream kryoBuffer;
+
+
+    @Setup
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
```diff
@@ -72,6 +72,22 @@ public void serializerPojo() throws Exception {
 		env.execute();
 	}
 
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkMiniBenchmarks.RECORDS_PER_INVOCATION)
+	public void serializerHeavyPojo() throws Exception {
```
AHeise: I'd add String to the name, like `serializerStringHeavyPojo`. We are not testing huge pojo's per se, but rather big strings in particular. Alternative: `serializerBigStringPojo`...

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
```diff
@@ -72,6 +72,22 @@ public void serializerPojo() throws Exception {
...
+		LocalStreamEnvironment env =
+				StreamExecutionEnvironment.createLocalEnvironment(4);
+		ExecutionConfig executionConfig = env.getConfig();
+		executionConfig.registerPojoType(MyPojo.class);
+		executionConfig.registerPojoType(MyOperation.class);
+
+		env.addSource(new HeavyPojoSource(RECORDS_PER_INVOCATION, 10))
```
AHeise: Dito name: `StringHeavyPojoSource`?

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
```diff
@@ -128,6 +144,49 @@ public void serializerRow() throws Exception {
...
+			super(numEvents, numKeys);
+		}
+
+		@Override
+		protected void init() {
+			super.init();
+			template = new MyPojo(
+					0,
+					makeString('a', 128),
+					new String[] {makeString('ж', 128), makeString('寿', 128)},
```
AHeise: Would it make any difference if the string are not replicated, but rather random? In particular what if characters of different UTF lengths are mixed?
I'm assuming that there should be no difference whatsoever, as the JVM cannot be so intelligent to optimize the cases. But still wanted to raise the question as it immediately popped into my mind.

AHeise (changes requested):
Thank you very much for the PR!

Code looks good. 1 question and 1 name change proposal. If that gets addressed, I'm happy to merge.

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
shuttie: Done

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
shuttie: done

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
shuttie: Nice idea, thanks. Done.

shuttie:
I've also added a set of more narrow benchmarks measuring pure serialization time, without job setup overhead. Currently it runs over different tastes of pojo (simple, avro, kryo) and different variants of strings (ascii, russian and chinese, 4-32-256 symbols) to showcase the impact of the FLINK-14246.

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
```diff
@@ -128,6 +146,50 @@ public void serializerRow() throws Exception {
...
+		public StringHeavyPojoSource(int numEvents, int numKeys) {
+			super(numEvents, numKeys);
+		}
+
+		@Override
+		protected void init() {
+			super.init();
+			template = new MyPojo(
+					0,
+					makeString(StringSerializationBenchmark.asciiChars, 128),
```
pnowojski: If we bump the string length to 1024, does that shows bigger difference for between and after the improvement? or even higher, like 4KB?  Or simplifying the PoJo to only a single/two 4,8 or 16KB strings?

It would be better to have this single benchmark. Adding more benchmarks (like `StringSerializationBenchmark` or `PojoSerializationBenchmark`) has to be carefully weighted as the trade off is longer benchmarking times.

Some partial solution might be to move those two `StringSerializationBenchmark` and `PojoSerializationBenchmark` benchmarks to `full` package.

AHeise: If we want to focus on the core of the contribution, `StringSerializationBenchmark` is the best way to see the impact. However, the continuously running benchmark suite is mostly for finding regressions and I doubt that we often change the string serialization, such that this benchmark will stay constant most of the time. Thus, I like the idea of moving that test to to `full`, so we can use it to drill down investigations if the serialization indeed worsens.

The same is true for `PojoSerializationBenchmark`, although it has a broader scope and could also show regressions in other serialization code paths. Nevertheless, both tests are superseded by this `SerializationFrameworkMiniBenchmarks`, which is enough to indicate issues in serialization and thus should be the only benchmark not in `full`. 

It would be awesome to just have one test, where we see a big impact by FLINK-14346. Imho the current 20% is a huge improvement, but as @pnowojski pointed out, it may be even more obvious if you could increase the size for this one benchmark to even some higher values.

shuttie:
@pnowojski see comments for [the original PR](https://github.com/apache/flink/pull/10358#issuecomment-561581868) for huge string benchmarks. TLDR: they are as good as for smaller strings.

@AHeise I've moved `StringSerializationBenchmark` and `PojoSerializationBenchmark` to the full package, so they won't be included into the normal benchmarking run.

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
pnowojski: @shuttie re-opening the issue.
> see comments for the original PR for huge string benchmarks. TLDR: they are as good as for smaller strings.

I wasn't concerned with improvement not being visible on huge strings. On the contrary. I was hoping that with larger strings, the performance improvements that you made will be even more visible (for example +50% instead of +20%), out-weighting other overheads of `SerializationFrameworkMiniBenchmarks#serializerStringHeavyPojo`, so testing even better for the string serialiser changes.

@AHeise agree, we can have more detailed benchmarks in `full` package even if tweak this benchmark to stress more string serialiser.

AHeise (approved):

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
AHeise: Okay then one last round to build on what @pnowojski . The full benchmarks are good as they are. The question is can we make the improvements in this one test `serializerStringHeavyPojo` more pronounced.
I'm currently thinking of simplifying the test by simply using a `LongStringSource` that only returns strings of size 1024. Then we would get rid of the constant tuple serialization overhead.

A rough sketch of the source:
```java
	/**
	 * Source emitting a long String.
	 */
	public static class LongStringSource extends BaseSourceWithKeyRange<String> {
		private static final long serialVersionUID = 3746240885982877398L;
		private String[] templates;

		public LongStringSource(int numEvents, int numKeys) {
			super(numEvents, numKeys);
		}

		@Override
		protected void init() {
			super.init();
			templates = new String[] {
				makeString(StringSerializationBenchmark.asciiChars, 1024),
				makeString(StringSerializationBenchmark.russianChars, 1024),
				makeString(StringSerializationBenchmark.chineseChars, 1024)
			};
		}

		private String makeString(char[] symbols, int length) {
			char[] buffer = new char[length];
			Random random = ThreadLocalRandom.current();
			Arrays.fill(buffer, symbols[random.nextInt(symbols.length)]);
			return new String(buffer);
		}

		@Override
		protected String getElement(int keyId) {
			return templates[keyId % templates.length];
		}
	}
```

AHeise (changes requested):
One last round: the benchmark could even more focus on the string. I sketched a solution where we completely drop the tuple around the string. I'd be fine to simply have the code change without new benchmarks (as they take quite a bit of time to run).
Before merging, we need to squash the commits into 1-2 single commits (1 is probably much easier, 2 could be divided into one addition to the SerializationFrameworkMiniBenchmark and one commit for the full benchmarks).

shuttie:
@AHeise @pnowojski I also suspect that current implementation of heavy-string-benchmark is not stressing the string serializer at all. See this flame graphs of this benchmark produced by the async-profiler:

With the FLINK-14346 applied:
flame graph with fix
Raw image: http://dfdx.me/images/flink-ss-flame.svg

Today's master (for some reason async-profiler swapped write/read parts on the graph, beware):
flame graph no fix
Raw image: http://dfdx.me/images/flink-ss-flame-nofix.svg

Main points are:
1. On the write path, it spends 90% of the time doing memory management with `DataOutputSerializer.write`. I've did initial discovery and looks like it's bound by `System.arraycopy` while copying the data out of the `StringSerializer` internal buffer. As arraycopy is a JVM intrinsic on x86_64, I may suspect that it's being bound by the memory throughput here, but it may need additional confirmation. (an interesting option will be to stop doing double buffering in the string serializer and write directly to the DataOutputSerializer buffer, but it seems to be like playing with fire on a cliff, while pointing the loaded gun in your mouth)
2. On the read path, it spends 40% of the time on the Kryo serializer (as `MyPojo` has an `Object` field, which cannot be serialized with PojoSerializer).

So in practice, the benchmark spends only ~20% of the time in the actual code of StringSerializer directly. Doing 4x speedup in there will increase overall throughput for about 15%, what we are actually observing in the benchmarks.

As an easy fix I agree with @AHeise that we can implement a more isolated source emitting giant strings, so there will be no additional overhead for POJOs, Kryo, arrays. And it will hit all the pain points of the current string serializer. I will update the PR when it's ready, hopefully today.

shuttie:
Benchmarking is full of mysteries. While playing with the string benchmark, I've lowered the parallelism from 4 to 1 and surprizingly got strong +10% improvement on throughtput for the current master code, without any changes. Doing the same run with `jmh perf` profiler validated the assumption that with 4 threads doing writes and 4 threads doing reads, we almost completely saturated the memory bus (dual channel DDR4-2667 on ryzen7 2700) and CPU spent most of the time being stuck on RAM access.

The results are showing a strong 2x improvement in throughput for the test job:
```
this-PR  SerializationFrameworkMiniBenchmarks.serializerHeavyString  thrpt   30  195.414 ± 8.203  ops/ms
master   SerializationFrameworkMiniBenchmarks.serializerHeavyString  thrpt   30  101.782 ± 2.047  ops/ms
```

2x improvement is nice, but you may also wonder why we had 4x-7x improvement on a pure string read-write microbenchmark, and only 2x on an integration-style one. My opinion (which is just an opinion without proper benchmarking results) is that with the full-blown job with intermediate buffering between source and sink, we do an object round-trip through the main memory (or L3 cache), as intermediate buffer cannot fit into the L1/L2. So we are still saturating the memory bus here.

In a new version of the PR:
1. I've implemented a new version of string-heavy source as @AHeise suggested.
2. The parallelism for this particular benchmark is ONE, so we do measure the actual serialization code, and not the RAM bandwidth.
3. Squashed all the commits into single one.

pnowojski:
@shuttie keep in mind that the [continuous benchmarking](http://codespeed.dak8s.net:8000/) machine is the one that we care the most and it has different hardware. Usually I observe better throughputs on it compared to my local machine.

But over all I think your benchmark looks good to me.

src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
shuttie: @pnowojski the main problem with the original benchmark code was using parallelism=4. It measured memory bandwidth instead of the actual code performance. See the last comment in this PR for the details.

AHeise (approved):

AHeise:
Again, thank you very much for your contribution. The main PR also looks good, but I'll wait with merging until tomorrow, so that we have one benchmark run with the old master.

PR #37, opened 2019-12-11
[FLINK-15103] Set number of network buffers to be the same as before FLIP-49.

xintongsong:
@pnowojski


pnowojski (approved):
LGTM, let's give it a try.

PR #38, opened 2019-12-11
[FLINK-15199] Fix compile error after FLINK-14926

pnowojski:


carp84 (approved):
+1, thanks for the fix @pnowojski !

pnowojski:
Verified locally - merging without waiting for a slot on the travis to unblock some release blocker. Will be tested by benchmarking CI anyway.

PR #39, opened 2019-12-12
[FLINK-15070] Add benchmarks for blocking partition

sunhaibotb:
Benchmarks three types of blocking partition with a simple job: uncompressed file, compressed file, uncompressed mmap.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/BlockingPartitionBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/BlockingPartitionBenchmark.java
@@ -1,0 +1,126 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.benchmark.functions.LongSource;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.NettyShuffleEnvironmentOptions;
+import org.apache.flink.runtime.jobgraph.ScheduleMode;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+import org.apache.flink.streaming.api.graph.StreamGraph;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.io.IOException;
+
+/**
+ * JMH throughput benchmark runner.
+ */
+@OperationsPerInvocation(value = BlockingPartitionBenchmark.RECORDS_PER_INVOCATION)
+public class BlockingPartitionBenchmark extends BenchmarkBase {
+
+	public static final int RECORDS_PER_INVOCATION = 15_000_000;
+
+	public static void main(String[] args)
+			throws RunnerException {
+		Options options = new OptionsBuilder()
+				.verbosity(VerboseMode.NORMAL)
+				.include(".*" + BlockingPartitionBenchmark.class.getSimpleName() + ".*")
+				.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void uncompressedFilePartition(UncompressedFileEnvironmentContext context) throws Exception {
+		executeBenchmark(context.env);
+	}
+
+	@Benchmark
+	public void compressedFilePartition(CompressedFileEnvironmentContext context) throws Exception {
+		executeBenchmark(context.env);
+	}
+
+	@Benchmark
+	public void uncompressedMmapPartition(UncompressedMmapEnvironmentContext context) throws Exception {
+		executeBenchmark(context.env);
+	}
+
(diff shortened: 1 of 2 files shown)


zhijiangW (approved):
Thanks for this contribution @sunhaibotb ! LGTM!

sunhaibotb:
Thanks for the review @zhijiangW .

pnowojski:
Could you share what are the results in those three different modes?

src/main/java/org/apache/flink/benchmark/BlockingPartitionBenchmark.java
```diff
@@ -0,0 +1,125 @@
...
+		streamGraph.setScheduleMode(ScheduleMode.LAZY_FROM_SOURCES_WITH_BATCH_SLOT_REQUEST);
+
+		env.execute(streamGraph);
+	}
+
+	/**
+	 * Setup for the benchmark(s).
+	 */
+	@State(Thread)
+	public static class BlockingPartitionEnvironmentContext {
```
pnowojski: could you deduplicate code by extending `FlinkEnvironmentContext`? It would be better to minimize number of places where for example number of buffers is configured. (there were some committed changes yesterday)

pnowojski:
Thanks for the update. Code LGTM. One remaining thing is:
> Could you share what are the results in those three different modes?

sunhaibotb:
Thanks for reviewing @pnowojski .

The benchmark results are:
```
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  3317.522 ±  92.648  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  5752.585 ± 182.796  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  5280.945 ±  88.173  ops/ms
```

pnowojski:
Thanks for the update! ~Nice results.~ Merging.

Ops, update. Why are the compressed partitions slower? The disks I/O is too fast on the benchmarking machine?

zhijiangW:
After discussed offline, the current default 1 parallelism and data character are not suitable for compression scenarios, so the extra compression cost overwhelms the benefits. 

Maybe we can construct two different scenarios by adjusting the parallelism or data. One is friendly for compression scenario and the other is not. Then we can measure the effects for both sides.

sunhaibotb:
Add more benchmark results:
```
NVMe SSD, 20 parallelism, sequence long:
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  2846.219 ± 170.731  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  2743.092 ± 260.190  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2485.105 ± 288.393  ops/ms

NVMe SSD, 20 parallelism, constant long (1234567890L):
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  3857.310 ± 395.241  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  3197.415 ± 306.892  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2542.286 ± 183.819  ops/ms

NVMe SSD, 1 parallelism, sequence long:
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  3134.444 ±  45.042  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  5738.641 ± 211.760  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  5212.091 ± 252.371  ops/ms

NVMe SSD, 1 parallelism, constant long (1234567890L):
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  5408.550 ± 194.175  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  5204.308 ± 156.812  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  5346.183 ± 151.527  ops/ms

SATA, 20 parallelism, sequence long:
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  2712.098 ± 173.706  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  3243.202 ± 348.518  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2668.959 ± 236.104  ops/ms

SATA, 20 parallelism, constant long (1234567890L):
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  3717.434 ± 338.845  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  3096.975 ± 315.963  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2721.369 ± 246.695  ops/ms

SATA, 1 parallelism, constant long (1234567890L):
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  3790.744 ± 336.315  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  3095.597 ± 330.904  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2651.462 ± 282.817  ops/ms

SATA, 1 parallelism, sequence long
Benchmark                                              Mode  Cnt     Score     Error   Units
BlockingPartitionBenchmark.compressedFilePartition    thrpt   30  2726.746 ± 154.262  ops/ms
BlockingPartitionBenchmark.uncompressedFilePartition  thrpt   30  3055.203 ± 328.738  ops/ms
BlockingPartitionBenchmark.uncompressedMmapPartition  thrpt   30  2624.319 ± 269.110  ops/ms
```

pnowojski:
parallelism 20 might be an overkill (benchmarking machine has 4 cpu cores), but maybe we can increase the parallelism to 4? Indeed parallelism of 1 is causing the reads/writes to be always sequential.

Also all in all, the main purpose of those compressed/uncompressed benchmarks is not to compare those two options, but to find performance regressions in the code.

One thing that we could do, is to actually mention somewhere this relation between parallelism and the compression impact in the docs? Maybe we could initiate a network stack performance tuning page [in the docs](https://ci.apache.org/projects/flink/flink-docs-release-1.9/dev/table/tuning/)?  We could start with just a couple of sentences when it might be worth enabling compression.

For bounded partitions when disk IO is the bottleneck:
- slow and small number of disks compared to number of CPU cores
- higher parallelism puts more pressure on the disks IO

For pipelined when network bit rate is the bottleneck:
- in rare cases when using low quality network

What do you think?

sunhaibotb:
> but maybe we can increase the parallelism to 4?

+1.  
I will create a new PR to change the parallelism to 4. In addition, the `CoreOptions.TMP_DIRS` config is not set in this PR, which makes the path of files always be `/tmp`. the `CoreOptions.TMP_DIRS` config will be set as the current working directory.

> Maybe we could initiate a network stack performance tuning page in the docs? We could start with just a couple of sentences when it might be worth enabling compression.

I think it's very good to add such document page.

PR #40, opened 2019-12-12
[FLINK-15222] Move state benchmark utils into core repository

carp84:
The changes on flink-benchmark side after the [flink PR](https://github.com/apache/flink/pull/10548) is in.


pnowojski:
Could we freeze this for at least couple of days? Merging this commit would add one extra commit that would need to be reverted if someone is benchmarking some performance regressions (and there are 3 ongoing investigations). 

If I forget about this, could you ping me on next Thursday/Friday?

carp84:
Makes sense, thanks for the careful consideration @pnowojski ! Let's get back to this later.

carp84:
@pnowojski travis is green and not sure whether this one is ready to go? thanks.

pnowojski:
As the performance regressions haven't been resolved yet (for example FLINK-15103) I would prefer to wait a bit more if that's not a problem?

pnowojski:
The performance analysis are almost done, so merging this.

carp84:
Thanks for review, track and merge this PR @pnowojski !

PR #41, opened 2019-12-13
[FLINK-15069][benchmark] Supplement the pipelined shuffle compression case for benchmark

zhijiangW:
We try to supplement the compression case for a typical pipelined scenario to measure the cost/benefit.

The benchmark comparison w/o compression for 1000 channels, 4 writers and 100ms flushing timeout. And we can see that the compression for pipelined shuffle has extra overhead if the network bandwidth is not the bottleneck.

StreamNetworkCompressionThroughputBenchmarkExecutor:  30824.915 ± 879.331  ops/ms
StreamNetworkThroughputBenchmarkExecutor:  32422.964 ± 933.416  ops/ms

### DIFF
--- a/src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputCompressedBenchmarkExecutor.java
+++ b/src/test/java/org/apache/flink/benchmark/StreamNetworkThroughputCompressedBenchmarkExecutor.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.NettyShuffleEnvironmentOptions;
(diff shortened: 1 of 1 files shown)


AHeise (approved):

src/test/java/org/apache/flink/benchmark/StreamNetworkCompressionThroughputBenchmarkExecutor.java
```diff
@@ -0,0 +1,84 @@
...
+		Options options = new OptionsBuilder()
+			.verbosity(VerboseMode.NORMAL)
+			.include(".*" + StreamNetworkCompressionThroughputBenchmarkExecutor.class.getSimpleName() + ".*")
+			.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void networkPipelinedCompressionThroughput(CompressionEnvironment context) throws Exception {
```
pnowojski: I think this name might be too long? 

What about just `networkThroughput` with a single value parameter `COMPRESSED`? In that case it would be also semi consistent with the [existing benchmarks](http://codespeed.dak8s.net:8000/changes/)? 

Or without the parameter, but just rename to `networkThroughputCompressed`?

Or to make it fully consistent, what about just adding `"1000,100ms,COMP"` parameter to the existing benchmark
`StreamNetworkThroughputBenchmarkExecutor.MultiEnvironment#channelsFlushTimeout`?

I guess any of those three variants would be fine.

pnowojski:
```
StreamNetworkCompressionThroughputBenchmarkExecutor: 30824.915 ± 879.331 ops/ms
StreamNetworkThroughputBenchmarkExecutor: 32422.964 ± 933.416 ops/ms
```
nice results. I wonder why blocking bounded partition has bigger regression :/ Maybe because here we are spreading the workload among more CPU cores/threads (separate receiver & sender workers + netty threads)

src/test/java/org/apache/flink/benchmark/StreamNetworkCompressionThroughputBenchmarkExecutor.java
zhijiangW: Considering the current compression option might be experimental, then I did not plan to mix it with existing formal `StreamNetworkThroughputBenchmarkExecutor`. 
I would take the second option to only rename the method atm.

zhijiangW:
Thanks for the reviews @pnowojski @AHeise !

I updated the codes to rename the long method. The current benchmark results fit the expectation, and we will further discuss the blocking compression results in the respective PR.

PR #42, opened 2019-12-17
[hotfix] Change the parallelism and the temp directory used to benchmark the blocking partition

sunhaibotb:
- Parallelism of 1 causes the reads/writes to be always sequential and only covers the case of one reader. More parallelism should be more suitable for finding performance regressions of the code. Considering that the benchmarking machine has 4 CPU cores, we set the parallelism to 4.
- Changes the temp directory to the current working directory.


pnowojski:
Could you describe (and copy the description in the PR description, commit description and inside the comment in the code) why have you changed the parallelism?

pnowojski:
@sunhaibotb ping about updating the descriptions.

sunhaibotb:
Thanks for reviewing. I've updated the PR description, commit description and code comment. @pnowojski

PR #44, opened 2019-12-18
[FLINK-15171] string serialization benchmark to use proper i/o buffer implementation as on SerializationFrameworkMiniBenchmarks

shuttie:
This is a follow-up to the upstream [FLINK-15171](https://github.com/apache/flink/pull/10529) PR.

Originally `PojoSerializationBenchmark` and `StringSerializationBenchmark` used simple jdk default `byte[]`-backed input-output streams and views to test the serialization performance. But this resulted in a severe performance mismatch with the e2e tests in `SerializationFrameworkMiniBenchmarks`: by default Flink is using own offheap-based implementation of memory operations, which has completely different performance characteristics.

So, for example, on `StringSerializationBenchmark` [FLINK-14346](https://issues.apache.org/jira/browse/FLINK-14346) implementation of string serialization heavily outperformed the original implementation, being faster up to 15 times. But on `SerializationFrameworkMiniBenchmarks` there was performance degradation instead.

In this PR we do the following improvements:
* we introduce an `OffheapInputWrapper` which is using exactly the same implementations of memory read operations as in default Flink with `HybridMemorySegment`
* we switch to `DataOutputSerializer` to match the default one used in Flink for memory write operations.
* both `StringSerializationBenchmark` and `PojoSerializationBenchmark` now use proper read-write primitives, matching the behavior of `SerializationFrameworkMiniBenchmarks`.

These improvements allowed to reliably reproduce the mysterious performance regression discussed in [FLINK-15171](https://github.com/apache/flink/pull/10529).

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/full/OffheapInputWrapper.java
+++ b/src/main/java/org/apache/flink/benchmark/full/OffheapInputWrapper.java
@@ -1,0 +1,32 @@
+package org.apache.flink.benchmark.full;
+
+import org.apache.flink.core.memory.DataInputView;
+import org.apache.flink.core.memory.MemorySegment;
+import org.apache.flink.core.memory.MemorySegmentFactory;
+import org.apache.flink.runtime.io.network.api.serialization.SpillingAdaptiveSpanningRecordDeserializer;
+import org.apache.flink.runtime.io.network.buffer.Buffer;
+import org.apache.flink.runtime.io.network.buffer.FreeingBufferRecycler;
+import org.apache.flink.runtime.io.network.buffer.NetworkBuffer;
+
+import java.lang.reflect.Field;
+
+public class OffheapInputWrapper {
+    public SpillingAdaptiveSpanningRecordDeserializer<?> reader;
+    public Buffer buffer;
+    public DataInputView dataInput;
+
+    public OffheapInputWrapper(byte[] initialPayload) throws Exception {
+        reader = new SpillingAdaptiveSpanningRecordDeserializer<>(new String[0]);
+        MemorySegment segment = MemorySegmentFactory.allocateUnpooledOffHeapMemory(initialPayload.length, this);
+        segment.put(0, initialPayload);
+        buffer = new NetworkBuffer(segment, FreeingBufferRecycler.INSTANCE, true, initialPayload.length);
+        Field nonSpanningWrapper = reader.getClass().getDeclaredField("nonSpanningWrapper");
+        nonSpanningWrapper.setAccessible(true);
+        dataInput = (DataInputView) nonSpanningWrapper.get(reader);
+    }
+
+    public void reset() throws Exception {
+        reader.setNextBuffer(buffer);
+    }
+
+}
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/flink/benchmark/full/PojoSerializationBenchmark.java
```diff
@@ -86,42 +86,69 @@ public static void main(String[] args)
     }
 
     @Benchmark
-    public byte[] writePojo() throws IOException {
-        return write(pojoSerializer, pojo);
+    @OperationsPerInvocation(INVOCATIONS)
```
pnowojski, line 89: nit: could this be moved to the top of the class?

src/main/java/org/apache/flink/benchmark/full/PojoSerializationBenchmark.java
```diff
@@ -86,42 +86,69 @@ public static void main(String[] args)
...
+    public int writePojo() throws IOException {
+        stream.pruneBuffer();
+        for (int i = 0; i < INVOCATIONS; i++) {
```
pnowojski, line 92: Could you motivate why do we need multiple invocations in a single benchmark? Because of potential costs of `stream.pruneBuffer();` or `pojoBuffer.reset();`? Are they measurable?

As it is, I would be a little be concerned what magic JIT can do after inlining `pojoSerializer.serialize(pojo, stream);`  and unrolling the loop. That might be desireable (vectorisation) but as we are invoking it over and over again with the same parameters, some other magic could yield false results.

src/main/java/org/apache/flink/benchmark/full/OffheapInputWrapper.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public OffheapInputWrapper(byte[] initialPayload) throws Exception {
+        reader = new SpillingAdaptiveSpanningRecordDeserializer<>(new String[0]);
+        MemorySegment segment = MemorySegmentFactory.allocateUnpooledOffHeapMemory(initialPayload.length, this);
+        segment.put(0, initialPayload);
+        buffer = new NetworkBuffer(segment, FreeingBufferRecycler.INSTANCE, true, initialPayload.length);
+        Field nonSpanningWrapper = reader.getClass().getDeclaredField("nonSpanningWrapper");
+        nonSpanningWrapper.setAccessible(true);
+        dataInput = (DataInputView) nonSpanningWrapper.get(reader);
+    }
```
pnowojski, line 26: There are a couple of issues here.

First, this relays on a non public Flink's API (`SpillingAdaptiveSpanningRecordDeserializer`), which means it can change at any point of time and brake the benchmark builds.

Secondly, this is using reflections to access a private field of an `Internal` class, which makes this benchmark even more fragile (and could cause compilation [issues like this recent one](https://issues.apache.org/jira/browse/FLINK-15199)).

Is there a way to avoid using internal classes? If there is really a no way, then it could be fixed similarly how https://github.com/dataArtisans/flink-benchmarks/pull/40 is refactoring the 15199 (take a look at `org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils` class).

Here in this case, the `OffheapInputWrapper` class should be defined and tested in `flink` repository, with some comment/annotation that it's being used by micro benchmarks. + code should be adjusted that the reflection is no longer needed.

pnowojski (changes requested):

Myasuka:
@shuttie this repo will be mitigated to https://github.com/apache/flink-benchmarks, please create this PR under that repo once [FLINK-17281](https://issues.apache.org/jira/browse/FLINK-17281) is completed.

PR #45, opened 2019-12-20
[FLINK-15340][network] Remove the executor of pipelined compression benchmark

zhijiangW:
In FLINK-15308, we removed the function of compression for pipelined case.
Accordingly we also need to remove the respective benchmark executor.


wsry:
Thanks @zhijiangW . LGTM.

wsry (approved):

PR #46, opened 2019-12-20
[hotfix] Remove useless `setCreateIfMissing`

Myasuka:
We don't need to call setCreateIfMissing explicitly since `RocksDBResourceContainer#getDbOptions` would add that. Moreover, every time we call `getDbOptions` we would initialize a native owned `DBOptions` once.


pnowojski:
@Myasuka could you open the PR against `flink` for that? There is a pending PR https://github.com/dataArtisans/flink-benchmarks/pull/40 that drops the class you are modifying in favour of using an equivalent in `flink` repository. 

If we merge this one here, it will probably create some conflict for https://github.com/dataArtisans/flink-benchmarks/pull/40.

Myasuka:
@pnowojski sorry, I forget to look up FLINK-15222 and once that merged, this hotfix PR would be useless. I'll close this PR now.

Issue #47, opened 2019-12-23
listState.update in #ListStateBenchmark#listAdd?


Jiayi-Liao:
https://github.com/dataArtisans/flink-benchmarks/blob/master/src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java#L119

This is introduced in [PR#13](https://github.com/dataArtisans/flink-benchmarks/pull/13/files).

Shouldn't this be `listState.add` ? 

Please correct me if I'm wrong. Thanks.

cc @carp84

carp84:
The existing `listAdd` benchmark is in accordance with the semantic of `add` benchmarks for `MapState` and `ValueState`, meaning the "insert value for new key" scenario.

However, since there's a specific `add` interface for `AppendingState`, I agree it's necessary to add another benchmark for `ListState`. How about we name the new benchmark as `listAppend`? Would you like to submit a PR for it? @buptljy Thanks.

Jiayi-Liao:
@carp84 Thanks for your explaination. 
+1 for adding a new benchmark for `ListState.add`. I'll submit a PR soon.

PR #48, opened 2019-12-24
[FLINK-16042] Add benchmark listAppend to test add function in AppendingState

Jiayi-Liao:
As discussed in [issue47](https://github.com/dataArtisans/flink-benchmarks/issues/47), to test the performance of `void add(IN value)` in `AppendingState`, we need to add a new Benchmark here.

The result looks like below, 
```
# Run complete. Total time: 00:07:51

Benchmark                      (backendType)   Mode  Cnt     Score    Error   Units
ListStateBenchmark.listAppend           HEAP  thrpt   30  3024.019 ± 36.616  ops/ms
ListStateBenchmark.listAppend        ROCKSDB  thrpt   30   315.488 ±  7.003  ops/ms
```
(Just to prove this works since I don't have a benchmark machine.)

### DIFF
--- a/src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
+++ b/src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
@@ -120,6 +120,12 @@ public class ListStateBenchmark extends StateBenchmarkBase {
     }
 
     @Benchmark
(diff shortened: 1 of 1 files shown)


carp84 (approved):
+1

PR #49, opened 2020-01-03
[FLINK-15103] Set number of network buffers to 1000

pnowojski:
With fewer number of buffers benchmarks should spend less time on initialisation and it should free up more memory for the benchmarks. 1000 buffers should be enough for all of the benchmarks, as they are executed on a single machine with low parallelism.


rkhachatryan (approved):
LGTM

PR #50, opened 2020-01-24
[FLINK-15748] Set explicit managed memory for RocksDBStateBenchmark

Myasuka:
After FLINK-15692, we set RocksDB to use managed memory by default. For `RocksDBStateBenchmark`, since we did not set any explicit managed memory, and it will use `16MB` which is not enough for performance test.

 I suggest to set the managed memory as 322122552 bytes, which is the default managed memory of 1GB TM with 1 slot.


pnowojski (approved):
@Myasuka I hope this will not increase some constant initialisation times?

Myasuka:
@pnowojski I only add that configuration to `RocksStateBackendBenchmark` and the initialisation time should stay the same with that when we enable memory limit by default. After we enable memory limit for RocksDB by default, the initialization time should be very light-weight compared with previous non-limit case.

PR #51, opened 2020-02-12
[FLINK-16019][benchmark] use NoRestart in CFRO benchmark

rkhachatryan:
There is an error currently in `ContinuousFileReaderOperator#dispose` which fails a single job run.
But the benchmark uses the default restart strategy which leads to indefinite restarts.
This PR changes restart strategy to NoRestart.

(CI failure after the 1st commit is the desired result (as long as there is an error in `ContinuousFileReaderOperator`))

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
@@ -19,6 +19,7 @@ package org.apache.flink.benchmark;
 
 import joptsimple.internal.Strings;
 import org.apache.flink.api.common.io.FileInputFormat;
+import org.apache.flink.api.common.restartstrategy.RestartStrategies;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.core.fs.FileInputSplit;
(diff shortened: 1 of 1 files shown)


pnowojski:
This still requires a fix in the master flink branch, right? And the failed travis run here is also caused by the not yet fixed master issue?

rkhachatryan:
Right, there should be at least two fixes: in benchmark and in the tested class itself.

src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -19,13 +19,15 @@
...
 import joptsimple.internal.Strings;
 import org.apache.flink.api.common.io.FileInputFormat;
+import org.apache.flink.api.common.restartstrategy.RestartStrategies;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.core.fs.FileInputSplit;
 import org.apache.flink.core.fs.Path;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
 import org.apache.flink.streaming.api.functions.source.ContinuousFileReaderOperator;
```
pnowojski: unused import?

rkhachatryan: ~The restart strategy was changed to prevent indefinite restarts in the future:~
~```env.setRestartStrategy(new RestartStrategies.NoRestartStrategyConfiguration());```~

pnowojski (approved):

PR #52, opened 2020-03-06
[FLINK-16459][benchmark] add Apache Thrift based serialization benchmarks

NicoK:
This PR adds benchmarks to `SerializationFrameworkAllBenchmarks` that are based on using Apache Thrift via Kryo.

Unfortunately, it is not possible to download the thrift executable from maven
central and use that to compile the data definition into a Java class. Since no
extra runtime dependencies are needed once that is done, we provide the
generated Java classes in the repository.

If you want to create updated Java Thrift classes, you need to install the
thrift binary with a matching version to the dependency inside the pom.xml or
override it and generate classes via

> mvn generate-sources -Pgenerate-thrift -Dthrift.version=0.13.0

or more dynamically:

> mvn generate-sources -Pgenerate-thrift -Dthrift.version=$(thrift --version | rev | cut -d' ' -f 1 | rev)

Be sure to use the same thrift version when compiling and running
tests/benchmarks as well to avoid potential conflicts.

Examplary Benchmark run:
```
Benchmark                                 Mode  Cnt     Score    Error   Units
serializerAvro                           thrpt   30   661.345 ± 30.075  ops/ms
serializerAvroReflect                    thrpt   30   114.142 ±  3.542  ops/ms
serializerHeavyString                    thrpt   30   210.036 ±  2.734  ops/ms
serializerKryo                           thrpt   30   306.214 ± 14.382  ops/ms
serializerKryoProtobuf                   thrpt   30   913.351 ± 24.035  ops/ms
serializerKryoThrift                     thrpt   30   348.893 ± 16.150  ops/ms
serializerKryoThriftWithoutRegistration  thrpt   30   357.520 ± 13.993  ops/ms
serializerKryoWithoutRegistration        thrpt   30   205.663 ±  9.902  ops/ms
serializerPojo                           thrpt   30   880.954 ± 32.082  ops/ms
serializerPojoWithoutRegistration        thrpt   30   868.786 ± 30.351  ops/ms
serializerRow                            thrpt   30  1162.336 ± 29.574  ops/ms
serializerTuple                          thrpt   30  1297.177 ± 30.248  ops/ms
```

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/52)


src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
```diff
@@ -85,4 +90,76 @@ public void serializerAvroReflect(FlinkEnvironmentContext context) throws Except
 
 		env.execute();
 	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkMiniBenchmarks.RECORDS_PER_INVOCATION)
+	public void serializerKryoThriftWithoutRegistration(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: this name is a bit lengthy

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
```diff
@@ -85,4 +90,76 @@ public void serializerAvroReflect(FlinkEnvironmentContext context) throws Except
...
+		env.addSource(new ThriftPojoSource(RECORDS_PER_INVOCATION, 10))
+				.rebalance()
+				.addSink(new DiscardingSink<>());
+
+		env.execute();
+	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = SerializationFrameworkMiniBenchmarks.RECORDS_PER_INVOCATION)
+	public void serializerKryoThrift(FlinkEnvironmentContext context) throws Exception {
```
pnowojski, line 96: You have put the benchmarks in `AllBenchmarks`, that are not executed automatically contrary to the `SerializationFrameworkMiniBenchmarks`. Was it intended?

src/main/resources/thrift/mypojo.thrift
```diff
@@ -0,0 +1,19 @@
+namespace java org.apache.flink.benchmark.thrift
```
pnowojski, line 16: can we add a comment on top of the thrift file explaining how to generate definitions (copy paste commit message)?

pnowojski (commented):
Thanks for the contribution :) Could you post example results of the new benchmarks?

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
NicoK: yes - I expect no changes in these benchmarks due to Flink code changes since they all go through Kryo.

pnowojski (approved):
LGTM % adding the comment and optionally if you can find out some shorter name for the benchmark (but I'm not blocking on this, as I can not propose anything better, apart of `serializerKryoThrift2`).

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
NicoK: yes, unfortunately...however, it matches the naming scheme of the other benchmarks. I'm not sure we should change that without changing the other ones as well, e.g. `NoReg` as the suffix. Since all `...WithoutRegistration` benchmarks are in the `SerializationFrameworkAllBenchmarks` class, this shouldn't be the problem, I guess. Tell me what you think...

NicoK:
Actually, I have to fix my initial commit due to a simple typo (see the fixup commit)...I'll redo the whole benchmarks on the machine that I used above, but on a different one, this is what changes with the hotfix:
```
Benchmark                                  Mode  Cnt    Score   Error   Units
serializerKryoThrift                      thrpt   30  162.844 ± 5.387  ops/ms
serializerKryoThrift2                     thrpt   30  318.193 ± 4.670  ops/ms
serializerKryoThrift2WithoutRegistration  thrpt   30  305.303 ± 6.561  ops/ms
serializerKryoThriftWithoutRegistration   thrpt   30  156.223 ± 4.235  ops/ms
```

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
NicoK: Or...since there is virtually no difference between the additional registration call and omitting it, we could completely remove this difference here and just use one test without the extra registration

NicoK:
Revised performance benchmark: (300000 records per invocation, Flink 1.11 snapshot)
```
Benchmark                                  Mode  Cnt     Score    Error   Units
serializerAvro                             thrpt   30   635.093 ± 25.948  ops/ms
serializerAvroReflect                      thrpt   30   112.972 ±  3.300  ops/ms
serializerHeavyString                      thrpt   30   207.010 ±  3.065  ops/ms
serializerKryo                             thrpt   30   304.366 ± 12.129  ops/ms
serializerKryoProtobuf                     thrpt   30   821.096 ± 20.180  ops/ms
serializerKryoProtobufWithoutRegistration  thrpt   30   831.122 ± 20.881  ops/ms
serializerKryoThrift                       thrpt   30   566.158 ± 19.034  ops/ms
serializerKryoThriftWithoutRegistration    thrpt   30   546.198 ± 15.343  ops/ms
serializerKryoWithoutRegistration          thrpt   30   190.028 ±  6.912  ops/ms
serializerPojo                             thrpt   30   820.420 ± 19.132  ops/ms
serializerPojoWithoutRegistration          thrpt   30   809.189 ± 24.638  ops/ms
serializerRow                              thrpt   30  1066.995 ± 24.468  ops/ms
serializerTuple                            thrpt   30  1165.520 ± 27.922  ops/ms
```

PR #54, opened 2020-03-31
[FLINK-16890] add Avro GenericRecord benchmark

NicoK:
This adds a benchmark for the third way of integrating with Avro serialization: via `GenericRecord`. While the current throughput is the same as with specific records, it may be important to have this to verify and to show how to integrate a `GenericRecord` correctly.

```
Benchmark               Mode  Cnt    Score    Error   Units
serializerAvro         thrpt   30  638.136 ± 23.087  ops/ms
serializerAvroGeneric  thrpt   30  638.280 ± 21.378  ops/ms
serializerAvroReflect  thrpt   30  120.000 ±  3.039  ops/ms
```

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/54)


src/main/java/org/apache/flink/benchmark/SerializationFrameworkMiniBenchmarks.java
```diff
@@ -158,7 +158,7 @@ public LongStringSource(int numEvents, int numKeys) {
 		}
 
 		@Override
-		protected void init() {
+		protected void init() throws Exception {
```
AHeise: In general, I'm not a big fan of Pokemon exception handling (gonna catch them all). Could we make it more specific (e.g. only IOException). 
I see the point that it makes implementing new sources much easier, so this comment is more of a nice-to-have.

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
```diff
@@ -126,6 +149,79 @@ public void serializerKryoProtobuf(FlinkEnvironmentContext context) throws Excep
...
+				return new Schema.Parser().parse(is);
+			}
+		}
+
+		@Override
+		protected void init() throws Exception {
+			super.init();
+
+			if (myPojoSchema == null) {
+				this.myPojoSchema = new Schema.Parser().parse(schemaString);
```
AHeise, line 187: I guess the main motivation for init() to throw exception is this schema parsing. But that should actually by handled in place and translated to an IllegalStateException (since the schema was provided in the ctor and just needed to be deserialized externally).

AHeise (commented):
Thank you for the contribution. I like having more benchmarks on serialization. My only request is to review the exception handling (no hard feelings if you decide to retain the current way).

src/main/java/org/apache/flink/benchmark/full/SerializationFrameworkAllBenchmarks.java
NicoK: oh, right...that code was different when I added the exceptions. I'll remove the exceptions again since there is no use for them any more

AHeise (approved):
LGTM % squash commits.

PR #55, opened 2020-04-08
[FLINK-17056][JMH] fix main methods running unrelated benchmarks

NicoK:
Each benchmark class is accompanied by an according `public static main (String[] args)` method which should run all benchmarks in that class. However, it just uses the class' simple name in a regexp like `".*<name>.*"` and may thus also match further classes that were not intended to run. An example for this is the `StreamNetworkThroughputBenchmarkExecutor` which also runs benchmarks from `DataSkewStreamNetworkThroughputBenchmarkExecutor`. Using the canonical name instead fixes that behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/55)


pnowojski (approved):
Just note, I recommend installing JMH plugin in Intellij and execute the benchmark class directly, instead of main class.

PR #56, opened 2020-04-08
 [FLINK-17057][network] add OpenSSL-based benchmarks

NicoK:
This PR and adds OpenSSL network benchmarks:

These can be executed either with dynamically-linked OpenSSL libraries or
statically-linked (and manually-provided) ones. Instructions are in README.md.

Current results on my PC:
- JDK 8:
```
Benchmark                                                   (channelsFlushTimeout)   Mode  Cnt      Score     Error   Units
StreamNetworkThroughputBenchmarkExecutor.networkThroughput          1000,100ms,SSL  thrpt   30  11760.342 ± 282.592  ops/ms
StreamNetworkThroughputBenchmarkExecutor.networkThroughput      1000,100ms,OpenSSL  thrpt   30  27889.053 ± 893.833  ops/ms
```
- JDK11:
```
Benchmark                                                   (channelsFlushTimeout)   Mode  Cnt      Score     Error   Units
StreamNetworkThroughputBenchmarkExecutor.networkThroughput          1000,100ms,SSL  thrpt   30  20628.476 ± 628.461  ops/ms
StreamNetworkThroughputBenchmarkExecutor.networkThroughput      1000,100ms,OpenSSL  thrpt   30  27289.926 ± 962.960  ops/ms
```

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/56)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,17 @@ We also support to run each benchmark once (with only one fork and one iteration
 mvn test -P test
 ```
 
+## Prerequisites
+
+The recent addition of OpenSSL-based benchmarks require one of two modes to be active:
+- dynamically-linked OpenSSL (default): this uses a dependency with native wrappers that are linked dynamically to your system's libraries. Depending on the installed OpenSSL version and OS, this may fail and you should try the next option:
+- statically-linked OpenSSL: this can be activated by `mvn -Dinclude-netty-tcnative-static` but requires `flink-shaded-netty-tcnative-static` in the version from `pom.xml` which can be generated from inside a corresponding flink-shaded source via:
+```
+mvn clean install -Pinclude-netty-tcnative-static -pl flink-shaded-netty-tcnative-static
+```
+
+If both options are not working, OpenSSL benchmarks will fail but that should not influence any other benchmarks.
+
 ## Code structure
 
 Recommended code structure is to define all benchmarks in [Apache Flink](https://github.com/apache/flink)
--- a/pom.xml
+++ b/pom.xml
@@ -31,6 +31,8 @@ under the License.
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<flink.version>1.11-SNAPSHOT</flink.version>
+		<flink.shaded.version>10.0</flink.shaded.version>
+		<netty.tcnative.version>2.0.25.Final</netty.tcnative.version>
 		<java.version>1.8</java.version>
 		<scala.binary.version>2.11</scala.binary.version>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -38,6 +38,17 @@ We also support to run each benchmark once (with only one fork and one iteration
...
+## Prerequisites
+
+The recent addition of OpenSSL-based benchmarks require one of two modes to be active:
+- dynamically-linked OpenSSL (default): this uses a dependency with native wrappers that are linked dynamically to your system's libraries. Depending on the installed OpenSSL version and OS, this may fail and you should try the next option:
+- statically-linked OpenSSL: this can be activated by `mvn -Dinclude-netty-tcnative-static` but requires `flink-shaded-netty-tcnative-static` in the version from `pom.xml` which can be generated from inside a corresponding flink-shaded source via:
+```
+mvn clean install -Pinclude-netty-tcnative-static -pl flink-shaded-netty-tcnative-static
+```
+
+If both options are not working, OpenSSL benchmarks will fail but that should not influence any other benchmarks.
```
pnowojski, line 50: Maybe instead of failing, default action should be not execute them? However I'm not sure how to achieve that with `@Params` from JMX :/

What's the error message that's going to be displayed? Maybe we could wrap it if it's not clear enough?

Do you know what will happen on the benchmarking machine?

NicoK: Depending on the exact error, the spawned JVM process may just die and we can't do anything to prevent that. Luckily, JMH just continues in that case, so the final results will just not include the failed benchmarks

pnowojski: As it is now, those benchmarks would fail on the Hetzner worker? Do you think it's worth spending time to try make them work? Would it involve some extra maintenance cost when upgrading Flink/netty?

NicoK: Good question, Debian- and RedHat-based distros should just work with our dynamically-linked OpenSSL bindings once [dependencies](https://netty.io/wiki/forked-tomcat-native.html#prerequisites-for-dynamically-linked-netty-tcnative) are installed: usually only libapr1 may be missing
If that is not usable, building the statically-linked netty-native is quite simple (as mentioned) since this basically only downloads appropriate artifacts. That should always work and here, we also do not need to look after licensing issues
Maintenance-wise, there could be overhead due to upgrades in the dynamically-linked version. Statically-linked bindings (when generated automatically for the appropriate Flink version) should come at no additional overhead.

pnowojski: Ok, thanks for the explanation @NicoK. I would leave it up to you if it's worth doing or not. I guess there is not that much of a value for us to benchmark OpenSSL, except of having some baseline numbers for recommendations Flink users, so I would be fine with NOT benchmarking it.

LGTM besides that

NicoK: I would actually like to keep these benchmarks as they may also show differences when we change configuration defaults that may have different effects on JDK SSL vs. OpenSSL.
It's just a single benchmark anyway - but if you don't want it in the per-commit runs, we could also move them to the ones you'd run manually

pnowojski: I meant that I wouldn't spend too much time in trying to make them work on Hetzner, but if you would like to try to do it, I'm +1 for it. The code for the benchmarks we can keep it has value, even if we are not using it, as long as it's not causing as extra maintenance :)

In other words, LGTM for the code changes, and it's up to you if you want to make sure OpenSSL is executing properly on Hetzner machine.

Myasuka:
Just remind whether this PR need to be merged before [FLINK-17281](https://issues.apache.org/jira/browse/FLINK-17281) is resolved.

NicoK:
@pnowojski if I see this correctly, you are approving this pr. Please express that via GitHub and then I'll merge (or you go ahead and do that after approving)

pnowojski (approved):

PR #57, opened 2020-04-22
[FLINK-17299][benchmarks] Add remote channel throughput benchmark with rapid checkpoint (1s checkpoint interval)

wsry:
To guarantee that the implementation of FLINK-16404 does not introduce any regressions for fast checkpoint scenario (1s checkpoint Interval and checkpoint complete in 1s), we introduce this benchmark. The following is the benchmark result:
```
Benchmark                                                                Mode  Cnt      Score      Error   Units
RemoteChannelThroughputBenchmark.remoteChannelWithCheckpointThroughput  thrpt   30  18377.411 ± 1774.382  ops/ms
```

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.benchmark.functions.LongSource;
+import org.apache.flink.runtime.minicluster.MiniCluster;
+import org.apache.flink.runtime.minicluster.MiniClusterConfiguration;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+import org.apache.flink.streaming.api.graph.StreamingJobGraphGenerator;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+@OperationsPerInvocation(value = RemoteChannelThroughputBenchmark.RECORDS_PER_INVOCATION)
+public class RemoteChannelThroughputBenchmark extends BenchmarkBase {
+    public static final int NUM_VERTICES = 3;
+    public static final int PARALLELISM = 4;
+    public static final int RECORDS_PER_SUBTASK = 10_000_000;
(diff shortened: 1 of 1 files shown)


zhijiangW (approved):
Thanks for this supplement @wsry !

It looks generally good on my side, a bit concern to reduce the parallelism to 4-5 for reducing the load for benchmark machine. Do you have other concerns @pnowojski ?

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @TearDown
+    public void tearDown() throws Exception {
+        if (miniCluster != null) {
+            miniCluster.close();
+        }
+    }
+
+    @Benchmark
+    public void remoteChannelWithCheckpointThroughput(FlinkEnvironmentContext context) throws Exception {
```
pnowojski: I would rename it to something like `remoteRebalanceMapSink` or maybe even just `remoteRebalance` to be more consistent with existing `mapSink` or `mapRebalanceMapSink` benchmarks?

`remoteDoubleRebalanceMapSink` or `remoteRebalanceMapRebalanceSink` would be more accurate, but I would expect this benchmark to be parametrised in the future with different checkpointing methods (unaligned? persistent communication channels?) and adding parameters would make the already long name way too long, so `remoteRebalance` probably would be the best name?

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+@OperationsPerInvocation(value = RemoteChannelThroughputBenchmark.RECORDS_PER_INVOCATION)
+public class RemoteChannelThroughputBenchmark extends BenchmarkBase {
+    public static final int NUM_VERTICES = 3;
+    public static final int PARALLELISM = 4;
+    public static final int RECORDS_PER_SUBTASK = 50_000_000;
+    public static final int RECORDS_PER_INVOCATION = RECORDS_PER_SUBTASK * PARALLELISM;
+
+    private static final long CHECKPOINT_INTERVAL_MS = 1000;
```
pnowojski: 1000? Other benchmarks are using 100ms. How long is single iteration? I would guess that with 1000ms there will be very few checkpoints, and one more or less could affect results significantly.

pnowojski:
Re the parallelism it's hard to say. For checkpointing actually number of channels matter a lot, so maybe parallelism of 4 is an ok value?

zhijiangW:
> Re the parallelism it's hard to say. For checkpointing actually number of channels matter a lot, so maybe parallelism of 4 is an ok value?

Yes, the setting of parallelism in first version was 10, then my above comment also suggested adjusting to 4-5 as a moderate option.

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
wsry: On my test machine, the benchmark method ```remoteChannelWithCheckpointThroughput``` takes more than 30s to finish, so there will be more than 30 checkpoints (I verified that by implementing ```CheckpointListener#notifyCheckpointComplete```).
I am not using 100ms because I think 100ms may be too short (equals to the default flush time) and is hardly used in practice.
I also tested 100ms checkpoint interval on my test machine, there is no regression, I reduce the checkpoint interval to 100ms and the records sent is also reduced which can make the test time shorter.

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
wsry: I changed the name to remoteRebalance.

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
pnowojski: @wsry 30 seconds for a single invocation to finish that's a really long time. That's ~25 minutes for this single benchmark.

Can not we try to bring this time down to <3s ? What are the results if you reduce `RECORDS_PER_SUBTASK` to 5M?

(Also that's the reason why other benchmarks are using 100ms checkpointing interval)

wsry: @pnowojski The 30s results come from 10 parallelism (I used 10 parallelism initially), not for 4 parallelism, sorry for the mistake. I ran the test with 4 parallelism and 10M records per parallelism on my test machine, it can already finish in 3s (about 2s). Do you think it is acceptable?

pnowojski: Yes, I think couple of seconds is fine.

In that case decreasing checkpointing interval might make more sense, but I'll leave it up to your judgement, as it could be also argued the opposite. Either way, LGTM.

wsry: Thanks for the review and comments. The current settings (100ms checkpoint interval and about 2s per invocation) make sense to me so I will keep the current settings.

zhijiangW:
Since @pnowojski also approved the current setting, merging it!

PR #60, opened 2020-05-14
[FLINK-17704][mvn] allow execution of arbitrary benchmarks via a property

NicoK:
This is based on #59 and adds a simple way of executing an arbitrary subset of benchmarks, e.g.

```
mvn -Dflink.version=1.11-SNAPSHOT exec:exec -Dbenchmarks="org.apache.flink.benchmark.StreamNetworkThroughputBenchmarkExecutor"
```

---
This change is [Reviewable](https://reviewable.io/reviews/dataartisans/flink-benchmarks/60)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,14 @@ benchmark suite to define runners to execute those test cases. You can execute t
 default benchmark suite (which takes ~1hour) at once:
 
 ```
-mvn -Dflink.version=1.5.0 clean install exec:exec
+mvn -Dflink.version=1.11-SNAPSHOT clean install exec:exec
 ```
 
 There is also a separate benchmark suit for state backend, and you can execute this suit (which takes ~1hour) using
 below command:
 
 ```
-mvn -Dflink.version=1.5.0 clean package exec:exec \
+mvn -Dflink.version=1.11-SNAPSHOT clean package exec:exec \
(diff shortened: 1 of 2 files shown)


pnowojski:
Wasn't there already a way to do it? It was written down in the readme:
https://github.com/dataArtisans/flink-benchmarks
> mvn -Dflink.version=1.5.0 clean package exec:exec \
 -Dexec.executable=java -Dexec.args="-jar target/benchmarks.jar -rf csv org.apache.flink.state.benchmark.*"

At least the README should be updated.

NicoK:
yes, the previously mentioned way was basically the same as executing "java -jar ..." yourself with all the knowledge required to run it (except for the classpath being added automatically). The new way should be much simpler and less error-prone in case of changes.

I updated the README and also changed the implementation to allow benchmarks that are excluded in the default profile.

To verify the profiles are working, I ran these commands and compared the resulting command lines, active profiles, etc:

```
mvn -Dflink.version=1.11-SNAPSHOT exec:exec clean package -Dbenchmarks="org.apache.flink.state.benchmark.ValueStateBenchmark.valueUpdate" -e -X
mvn -Dflink.version=1.11-SNAPSHOT exec:exec clean package -e -X
mvn help:active-profiles -P generate-thrift
mvn clean package -P generate-thrift
```

pnowojski (approved):

PR #63, opened 2020-05-29
[FLINK-16057] Don't end input in ContinuousFileReaderOperator Benchmark

rkhachatryan:
Currently, END_OF_INPUT is sent once all splits were emitted.
Thus, all subsequent reads in ContinuousFileReaderOperator are made in
CLOSING state in a simple loop (MailboxExecutor.isIdle is always true).
Which is not a scenario this benchmark is intended for.

This change makes source to wait until all elements reach sink.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
@@ -24,18 +24,24 @@ import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.core.fs.FileInputSplit;
 import org.apache.flink.core.fs.Path;
+import org.apache.flink.core.testutils.OneShotLatch;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/flink/benchmark/ContinuousFileReaderOperatorBenchmark.java
```diff
@@ -45,6 +51,7 @@
 
     private static final TimestampedFileInputSplit SPLIT = new TimestampedFileInputSplit(0, 0, new Path("."), 0, 0, new String[]{});
     private static final String LINE = Strings.repeat('0', 10);
+    private static OneShotLatch TARGET_COUNT_REACHED_LATCH = new OneShotLatch();
```
pnowojski, line 57: Could you copy the explanation from the commit message into a comment here, describing the intention behind this latch?

pnowojski (approved):

pnowojski:
Could you take a look at the build failure?

rkhachatryan:
Thanks for the feedback.
I've fixed the failure on travis and addressed the comment, merging.

